PR #11, opened 2021-01-13
Avoid note about non-standard file in root

maurolepore:
Fix R CMD note:

```
❯ checking top-level files ... NOTE
  Non-standard file/directory found at top level:
    ‘README.Rmd’
```


MonikaFu (approved):
Thank you!

LGTM

PR #14, opened 2021-01-13
Polish DESCRIPTION

maurolepore:
DESCRIPTION is a picky file. It is a very special file because it
defines an R package. CRAN is very picky about how to write it;
for example, the title must use Title Case not Sentense case and
not end in dot ".". Also, they try to avoid the use of the word
'package' and they ask for package names.

This convensions are totally idiosyncratic and not worth learning
unless you submit to CRAN. But as I'm here, why not try meet those
CRAN policies?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,19 +1,23 @@
 Package: r2dii.ggplot
-Title: Creates standard 2DII plots using ggplot
+Title: Create Standard Static 2DII Plots
 Version: 0.0.0.9000
 Authors@R: 
     person(given = "Monika",
            family = "Furdyna",
            role = c("aut", "cre"),
            email = "monika.furdyna@gmail.com",
-           comment = c(ORCID = "YOUR-ORCID-ID"))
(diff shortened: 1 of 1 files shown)


DESCRIPTION
```diff
@@ -1,19 +1,23 @@
...
+Title: Create Standard Static 2DII Plots
 Version: 0.0.0.9000
 Authors@R: 
     person(given = "Monika",
            family = "Furdyna",
            role = c("aut", "cre"),
            email = "monika.furdyna@gmail.com",
-           comment = c(ORCID = "YOUR-ORCID-ID"))
-Description: This package allows the user to create standardized 2DII graphis, like trajectory charts or technology mix charts and outputs them as ggplot objects. It also contains data processing functions that prepare the standard bank's data to be an input for the plotting functions.
+           comment = c(ORCID = "FIXME"))
```
maurolepore, line 9: Remember to add your ORCID

MonikaFu (approved):
Thank you! LGTM

PR #15, opened 2021-01-13
Add workflows for GitHub actions

maurolepore:
GitHub actions can run code on our behalf. Here I set it up to run
workflows that other r2dii packages use. They come from
https://github.com/r-lib/actions/blob/master/examples and can be
used easily with some form of `usethis::use_github_action*()`


MonikaFu (approved):
LGTM after @maurolepore gave me a short introduction into the idea of GitHub actions

PR #17, opened 2021-01-14
Remove always_allow_html

maurolepore:


cjyetman:
out of curiosity, was this causing a specific problem?

MonikaFu:
> out of curiosity, was this causing a specific problem?

@cjyetman - from my understanding - not. It is more from the cleaning perspective -  if not necessary it should be removed.

MonikaFu (approved):
LGTM

MonikaFu (approved):
LGTM

maurolepore:
> out of curiosity, was this causing a specific problem?
No, not at all.

cjyetman:
> > out of curiosity, was this causing a specific problem?
> 
> No, not at all.

cool... then my repo is safe 😉

maurolepore:
By now you surely know more than me but I suspect that option is a must for interactive output.

PR #18, opened 2021-01-15
Reorder scenarios and colours of 'green' technologies

### LINKED ISSUE #5: `plot_trajectory_chart` outputs unexpectedly for low-carbon technologies
When plotting renewables, I would expect dark green to occur as the production increases. The results of the reprex below surprise me: 

---

Brief description of the problem

``` r
library(r2dii.ggplot)

data <- get_example_data()
processed_data <- process_input_data(data)

data_trajectory <- filter_data_for_trajectory_chart(
  processed_data,
  sector = "power", 
  technology = "renewablescap",
  region = "global",
  scenario_source = "demo_2020",
  value_name = "production", 
  end_year = 2025,
  normalize_to_start_year = FALSE
)

scenario_specs <- data.frame(scenario = c("sds","sps","cps"),
                             color = c("#9CAB7C","#FFFFCC","#FDE291"),
                             label = c("SDS","STEPS","CPS"))
worstColor <- "#E07B73"
mainLineMetric <- data.frame(metric = "projected", label = "Portfolio")
additionalLineMetrics <- data.frame()

plot <- plot_trajectory_chart(data_trajectory,
                              plotTitle = "Production trajectory of Renewables Capacity technology \n in the Power sector",
                              xTitle = "Year", yTitle = "Production (MW)",
                              annotateData = FALSE,
                              scenario_specs,worstColor,
                              mainLineMetric, additionalLineMetrics)
#> Selecting by value_low

plot
```

<sup>Created on 2021-01-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

MonikaFu:
/style

Closes #5 

In this PR I modify the code around the trajectory chart so that it reverts the order of scenarios and colours in case the technology which we are plotting is 'green' (based on `green_or_brown` data set from `r2dii.data` package).

Testing:
Charts for "oilcap" (brown) and "renewablescap" (green) from the example data output correct plots.


maurolepore:
Thanks! Before I see the code, I addes a few [commits](https://github.com/2DegreesInvesting/r2dii.ggplot/pull/18/commits) for stuff that is kind of standard in r2dii packages (e.g. wrap help-file documentation at 80 characters).

R/plot_2diicharts.R
```diff
@@ -31,20 +30,17 @@ create_general_plot_with_default_settings <- function() {
...
 #' The function returns a ggplot object containing a stacked bar chart showing a technology mix for different categories (portfolio, scenario, benchmark, etc.)
 #'
-#' @import ggplot2
-#' @import dplyr
 #' @export
 
 plot_trajectory_chart <- function(data, plotTitle = "", xTitle = "", yTitle = "", annotateData = FALSE,
-                                  scenario_specs, worstColor = "#E07B73", mainLineMetric,
+                                  scenario_specs_good_to_bad, mainLineMetric,
                                   additionalLineMetrics = data.frame()) {
```
maurolepore, line 44: It's best to stick to either camelCase or snake_case. r2dii packages use snake_case because they generally follow the tidyverse style guide.

R/plot_2diicharts.R
```diff
@@ -67,36 +63,58 @@ plot_trajectory_chart <- function(data, plotTitle = "", xTitle = "", yTitle = ""
...
+
+    data_scenarios <- data %>%
+      filter(.data$metric_type == "scenario") %>%
+      select(.data$year, .data$metric, .data$value)
+
+    data_scenarios <- rbind(data_scenarios, data_worse_than_scenarios) %>%
+      group_by(.data$year) %>%
+      arrange(.data$year, factor(.data$metric, levels = scenario_specs$scenario)) %>%
+      mutate(value_low = dplyr::lag(.data$value, n = 1, default = lower_area_border))
+  } else if (tech_green_or_brown == "green") {
```
maurolepore, line 86: Conditions on a string may be simplified with `switch()`.

R/plot_2diicharts.R
```diff
@@ -67,36 +63,58 @@ plot_trajectory_chart <- function(data, plotTitle = "", xTitle = "", yTitle = ""
...
+  }
 
   for (i in 1:length(scenario_specs$scenario)) {
     scen <- scenario_specs$scenario[i]
     color <- scenario_specs$color[i]
     data_scen <- data_scenarios %>% filter(.data$metric == scen)
     p_trajectory <- p_trajectory +
-      geom_ribbon(data = data_scen, aes(ymin = .data$value_low, ymax = .data$value, x = year, group = 1), fill = color, alpha = 0.75) +
-      geom_line(data = data_scen, aes(x = year, y = .data$value), color = color)
+      geom_ribbon(data = data_scen, aes(ymin = .data$value_low, ymax = .data$value, x = year, group = 1), fill = color, alpha = 0.75)
```
maurolepore, line 107: Best to wrap at 80 characters so we don't need to scroll right.

R/plot_2diicharts.R
```diff
@@ -67,36 +63,58 @@ plot_trajectory_chart <- function(data, plotTitle = "", xTitle = "", yTitle = ""
...
+  data_worse_than_scenarios <- data.frame(year)
 
-  p_trajectory <- p_trajectory +
-    geom_ribbon(
-      data = data_worst_than_scenarios,
-      aes(ymin = .data$value_low, ymax = .data$value, x = year, group = 1),
-      fill = worstColor, alpha = 0.75
-    )
+  green_or_brown <- r2dii.data::green_or_brown
+  tech_green_or_brown <- green_or_brown[green_or_brown$technology == data$technology[1], ]$green_or_brown
```
maurolepore, line 70: Filtering with `[` is totally valid but we are importing and using `dplyr::filter()` so may be reuse it for consistency.

maurolepore (approved):
Thanks

R/data_processing.R
```diff
@@ -46,7 +48,7 @@ filter_data_for_trajectory_chart <- function(data, sector, technology,
     filter(.data$region == !!region) %>%
     filter(.data$scenario_source == !!scenario_source) %>%
     filter(.data$year <= end_year) %>%
-    select(.data$year, .data$metric_type, .data$metric, value = !!value_name)
+    select(.data$year, .data$metric_type, .data$metric, .data$technology, value = !!value_name)
 
   if (normalize_to_start_year) {
```
jdhoffa, line 53: NIT: I think a similar thing could be achieved tidy-style with something like this: 
``` r
data_filtered %>%
  # you may need to change this line to group by the necessary columns
  group_by_at(setdiff(names(data_filtered), "year")) %>%  
  arrange(.data$year) %>% 
  mutate(value = .data$value / first(.data$value)) 
```
(I didn't test this so don't take my word, but I think something like that could work, and you wouldn't need to worry about `technology.x` or `year.y` and whatnot.)

R/imports.R
```diff
@@ -0,0 +1,4 @@
+#' @import ggplot2
```
jdhoffa, line 1: Comment: It's good to only import the functions from any given package that will actually be used, as is done with `dplyr` below. But I think since this package depends so heavily on `ggplot2` that this is probably totally fine.

R/plot_2diicharts.R
jdhoffa: +1 for this. 
There's a couple different places where camelCase is used. We should probably address this in the coding best practices sessions.

jdhoffa (approved):
Some non-blocking/ nit-picky comments, but the functionality looks good to me!
@MonikaFu have a look at the comments if you're curious/ have time, otherwise go ahead and merge if you just want it done :-) The plots lgtm

R/plot_2diicharts.R
MonikaFu: good point. Will fix

MonikaFu: I like camelCase btw 😆 but understand that snake_case is more common in R

R/plot_2diicharts.R
MonikaFu: it's not fully clear to me how I would do it so I will ignore this for now..

MonikaFu:
> /style

@maurolepore - I am not sure if this is working. I don't see any commits coming from it

R/data_processing.R
MonikaFu: thanks! Will try it in another PR as there is bug in this part of the code anyway

R/plot_2diicharts.R
maurolepore: Appart from `?switch()`, I like a talk by Jenny Bryan: [Code smells and feels](https://github.com/jennybc/code-smells-and-feels) -- here is the bit about `switch()`.
https://youtu.be/7oyiPBjLAWY?t=2395

maurolepore:
I suspect `/style` must the the firs and only text in the comment.

PR #22, opened 2021-01-19
Add 2dii colour palette to the package

### LINKED ISSUE #8: Add standard 2DII colour palette to the package
2DII colour palette should be added to the package so that it can be used for some of the line and bar plots.

MonikaFu:
closes #8 

In this PR I add a function that returns a data frame with 2DII colour palette. 

Also: some styling has been added (using `usethis::use_tidy_style()`) to plotting and data processing functions.


R/data_processing.R
```diff
@@ -90,8 +91,9 @@ filter_data_for_techmix_chart <- function(data, sector, years,
...
     mutate(
       metric_type = paste0(.data$metric_type, "_", as.character(.data$year))
-      ) %>%
+    ) %>%
     select(.data$technology, .data$metric_type, .data$metric,
-           value = !!value_name)
-    return(data_filtered)
+      value = !!value_name
+    )
+  return(data_filtered)
```
maurolepore, line 98: Nit: This is not the focus of this PR but gives me the chance to highlight a style detail. The tidyverse style guide recomends to reserve `return()` only to return early -- so in this case you would return `data_filtered` instead of `return(data_filtered)`.

R/plot_2diicharts.R
```diff
@@ -75,7 +80,8 @@ plot_trajectory_chart <- function(data, plot_title = "", x_title = "",
 
   green_or_brown <- r2dii.data::green_or_brown
   tech_green_or_brown <- green_or_brown[
-    green_or_brown$technology == data$technology[1], ]$green_or_brown
+    green_or_brown$technology == data$technology[1],
+  ]$green_or_brown
```
maurolepore, line 84: Nit: Maybe an opportunity to use `dplyr::fileter()` for consistency.

R/plot_2diicharts.R
```diff
@@ -313,3 +341,31 @@ get_sector_colors <- function(sector) {
 
   return(colors)
 }
+
+get_2dii_colours <- function() {
```
maurolepore, line 345: The prefix `get_*()` prefix  suggests this is a simpler helper to pull a piece of data. Instead, what I see here is a constructor, and the prefix `new_*()` seems more appropriate. 

Do we really need to construct the object? If so, I would [try use vtrs](https://vctrs.r-lib.org/articles/s3-vector.html#implementing-a-vctrs-s3-class-in-a-package-1). If no, consider this alternative:

``` r
get_2dii_colours <- function() {
  tibble::tribble(
           ~label, ~colour_hex,
      "dark_blue",   "#1b324f",
          "green",   "#00c082",
         "orange",   "#ff9623",
    "dark_purple",   "#574099",
         "yellow",   "#f2e06e",
      "soft_blue",   "#78c4d6",
       "ruby_red",   "#a63d57",
           "grey",   "#d0d7e1",
     "moss_green",   "#4a5e54"
    )
}

get_2dii_colours()
#> # A tibble: 9 x 2
#>   label       colour_hex
#>   <chr>       <chr>     
#> 1 dark_blue   #1b324f   
#> 2 green       #00c082   
#> 3 orange      #ff9623   
#> 4 dark_purple #574099   
#> 5 yellow      #f2e06e   
#> 6 soft_blue   #78c4d6   
#> 7 ruby_red    #a63d57   
#> 8 grey        #d0d7e1   
#> 9 moss_green  #4a5e54
```

<sup>Created on 2021-01-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

maurolepore (approved):
LGTM but please consider the this alternative implementation:

``` r
get_2dii_colours <- function() {
  tibble::tribble(
           ~label, ~colour_hex,
      "dark_blue",   "#1b324f",
          "green",   "#00c082",
         "orange",   "#ff9623",
    "dark_purple",   "#574099",
         "yellow",   "#f2e06e",
      "soft_blue",   "#78c4d6",
       "ruby_red",   "#a63d57",
           "grey",   "#d0d7e1",
     "moss_green",   "#4a5e54"
    )
}

get_2dii_colours()
#> # A tibble: 9 x 2
#>   label       colour_hex
#>   <chr>       <chr>     
#> 1 dark_blue   #1b324f   
#> 2 green       #00c082   
#> 3 orange      #ff9623   
#> 4 dark_purple #574099   
#> 5 yellow      #f2e06e   
#> 6 soft_blue   #78c4d6   
#> 7 ruby_red    #a63d57   
#> 8 grey        #d0d7e1   
#> 9 moss_green  #4a5e54
```

<sup>Created on 2021-01-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/plot_2diicharts.R
MonikaFu: I don't do it here because I actually want to get a string returned (which is under `$green_or_brown`) and when using `dplyr::filter()` I get a data frame back. If you say that I should still use it and get my string in the next step - will do.

cjyetman: not arguing for the `tidyverse` version, but I believe the `tidyverse` way would be to add `%>% dplyr::pull(green_or_brown)` to get the string/vector

cjyetman: in the `base`-y way, I prefer to extract the vector first and then do the subsetting on that, like...
``` r
green_or_brown$green_or_brown[green_or_brown$technology == data$technology[1]]
```

I would say `data$technology[1]` is questionable... should it be `unique(technology)[[1]]`? Is there a good reason why you only/always need just the first element?

cjyetman: also, the way that `r2dii.data::green_or_brown` is structured, it seems like you should be matching `sector` AND `technology`, because there are duplicate values in just `r2dii.data::green_or_brown$technology`

MonikaFu: they are all the same (or at least should be) - so that is why I take the first. It is a matter of getting 1 element instead of a vector.

R/plot_2diicharts.R
```diff
@@ -242,8 +268,10 @@ plot_techmix_chart <- function(data, plot_title = "", show_legend = TRUE,
   if (show_legend) {
     p_techmix <- p_techmix +
       theme(legend.position = "bottom") +
-      theme(legend.text = element_text(family = "Helvetica", size = 9,
-                                       margin = margin(5, 5, 5, 5))) +
+      theme(legend.text = element_text(
+        family = "Helvetica", size = 9,
```
cjyetman, line 272: careful with fonts.... not sure if `"Helvetica"` is portable across OSs

R/plot_2diicharts.R
cjyetman: personally, I would prefer just an object `2dii_colours` that is the data frame

colors/2dii_colour_palette.csv
```diff
@@ -0,0 +1,10 @@
+label,colour_hex
+dark_blue, #1b324f
+green, #00c082
+orange, #ff9623
+dark_purple, #574099
+yellow, #f2e06e
+soft_blue, #78c4d6
+ruby_red, #a63d57
+grey, #d0d7e1
+moss_green, #4a5e54
```
cjyetman, line 10: I like the idea of this CSV being the master copy of the color values, and adding a sysdata.R file that automatically reads it in and adds it as an data object that's available in the package. There are examples of this in other 2dii packages.

cjyetman (approved):

R/plot_2diicharts.R
cjyetman: to clarify a bit.... I don't imagine a situation where the user wants a "new" copy of this information (which implies they would want to modify it before using it for some reason?), they just want the information

MonikaFu: I imagine a situation when they would like to modify it, that is why I leave it a bit more flexible for now. To be determined in user tests.

colors/2dii_colour_palette.csv
MonikaFu: had a discussion about it with @maurolepore and he is of an opinion that the .csv adds additional level of complexity and for small dataframes it is more 'clean' to keep their definition directly in the code. That being said - this is WOP and I will wait with more input from the users before I decide what to do with the objects/data sets with colour specifications. There are many choices here and hard to predict which one is best.

R/plot_2diicharts.R
MonikaFu: aren't the checks that @maurolepore implemented on different machines with different OSs handling that issue?

maurolepore: It's be great to write unit tests that somehow confirms the desired font is available.Our gh-actions workflow would run such a test during R CMD check, and alert us if the font is unavailable in some platform.

Can anyone think of what that test would look like?

cjyetman: @MonikaFu the CI tests that currently run, as far as I know, only test if this repo can be built Into a pkg

PR #30, opened 2021-01-25
Remove example_script.R

### LINKED ISSUE #19: include a 2dii custom ggplot theme
Include a 2dii custom ggplot theme that could combine official 2dii color palette/s, official 2dii font specs, and other data viz guidances from @MonikaFu 

Such a custom ggplot theme could then be applied to any custom ggplot that a 2dii user may be creating.

see this package for some cool, complex examples: https://github.com/hrbrmstr/hrbrthemes

I believe @vintented may have made something like this before, or was that an Rmarkdown theme?

### LINKED ISSUE #29: Remove `example_script.R`
It looks as though the contents of `example_script.R` are also included in `README.Rmd`. I like the idea of having it in the `README`, as the examples will be run every time we knit the `.md`, and we can see directly on the homepage on github if the examples are running/ producing the plots we expect. 

Also, doubling the same script twice doesn't follow the DRY principle, and you risk having them fall out of sync easily. 

Relates to #25

MonikaFu:
Closes #29 

In this PR I remove the `example_script.R` from the repo and add it to the `.gitignore` file (I am going to use it locally for testing the functions).


maurolepore (approved):

jdhoffa (approved):
lgtm, Although I think this closes #29 instead of #19 ;-)

MonikaFu:
> lgtm, Although I think this closes #29 instead of #19 ;-)

you're right 😃

Issue #10, opened 2021-01-13
Add meta report charts to the package


MonikaFu:
Charts used in the meta report should be added to the package as well. Use the code in [meta_report repo](https://github.com/2DegreesInvesting/meta_report) as a starting point.

- [x] assets coverage
- [x] climate-relevant sectors coverage
- [x] climate-relevant sectors mix
- [x] distribution chart
- [x] bubble chart
- [x] map
- [ ] emission intensity

MonikaFu:
Moved to: [`r2dii.plot.internal`](https://github.com/2DegreesInvesting/r2dii.plot.internal/issues/5)

PR #34, opened 2021-01-29
Add relevant sectors chart

MonikaFu:
Related to #10 

In this PR I add a meta-report chart displaying PACTA sectors coverage to the package. 

image

In particular I:
* Add a data processing function for '_overview_portfolio.rda' dataset
* Add a plotting function displaying the information in a form of 'small multiples' for Equity and Bond separately.


cjyetman:
FYI, tests on R v3.5 and R v3.4 fail because
> Error: Package required but not available: ‘ggpubr’
> ERROR: dependencies ‘cowplot’, ‘rstatix’ are not available for package ‘ggpubr’

you may have a dependency on R v3.6+ now

MonikaFu:
Thanks for pointing it out @cjyetman . Any advice on how to proceed? Maybe @maurolepore ?

cjyetman:
probably need to specify that in DESCRIPTION and probably then remove those tests from the GitHub action that's running CI

jdhoffa (changes requested):
Could you add an example into @examples of `prepare_for_pacta_sectors_chart` and `plot_metareport_pacta_sectors`? It will make it easier to test the output, and also is helpful to have that documentation for future users. 

(Something like `get_example_data() %>% prepare_for_pacta_sectors_chart() %>% plot_metareport_pacta_sectors()`)

cjyetman:
I would wait for @maurolepore to comment though.... 
[ggpubr](https://CRAN.R-project.org/package=ggpubr) officially depends on v3.1.0
[cowplot](https://CRAN.R-project.org/package=cowplot) officially depends on v3.5.0
[rstatix](https://CRAN.R-project.org/package= rstatix) officially depends on v3.3.0

so honestly, I don't know why their installs are failing on the v3.5 test, though the v3.4 may obviously be because it can't install cowplot

MonikaFu:
> Could you add an example into @examples of `prepare_for_pacta_sectors_chart` and `plot_metareport_pacta_sectors`? It will make it easier to test the output, and also is helpful to have that documentation for future users.
> 
> (Something like `get_example_data() %>% prepare_for_pacta_sectors_chart() %>% plot_metareport_pacta_sectors()`)

@jdhoffa at the moment I can't because I do not have 'toy' meta report data. I could work on creating it but I thought that it might not be of highest priority at the moment.

cjyetman:
I think the dependency path is....
ggpubr->rstatix->car->lme4-> nloptr->A system installation of NLopt

which results in...
> ERROR: configuration failed for package ‘nloptr’
because it fails to make the underlying system dependencies (nlopt and its dependencies)

Welcome to dependency hell! 🔥

maurolepore:
Thanks CJ for noticing the dependency issues. I just pushed a commit that removes the broken checks on CI. I expect to re-enter the green zone. Eventually we can work to support older versions of R, but at this stage the priority seems to be to get software to beta users quickly so we can learn what we really want to build. 

Feel free to revert if this solution is not what you want.

MonikaFu:
> Thanks CJ for noticing the dependency issues. I just pushed a commit that removes the broken checks on CI. I expect to re-enter the green zone. Eventually we can work to support older versions of R, but at this stage the priority seems to be to get software to beta users quickly so we can learn what we really want to build.
> 
> Feel free to revert if this solution is not what you want.

Great! Thanks a lot @maurolepore

MonikaFu:
@jdhoffa in commit 012c9e0 I added some code example of how one would set up the parameters and run the charts code, assuming that they have the right data. I hope that helps for now.

maurolepore:
I highlight that https://github.com/2DegreesInvesting/r2dii.ggplot/issues/25 will remind us to transform the examples in README into examples in the helpfiles. As Jackson's suggestion won't fall through the cracks, in my opinion, you are good to go.

jdhoffa (approved):
sorry for the delay! I'm satisfied with this solution :-) 
lgtm

PR #36, opened 2021-02-01
Add PACTA sectors mix chart

MonikaFu:
Related to #10 

In this PR I add another meta report chart displaying the percentage investment in different PACTA sectors to the package. 

image

In particular I:

* Add a data processing function for '_overview_portfolio.rda' dataset
* Add a plotting function displaying the information in a form of 'small multiples' for Equity and Bond separately.


R/plot_2diicharts.R
```diff
@@ -119,7 +119,8 @@ plot_trajectory <- function(data, plot_title = "", x_title = "",
       ))
   } else if (tech_green_or_brown == "green") {
     scenario_specs <- scenario_specs_good_to_bad[
-      nrow(scenario_specs_good_to_bad):1, ]
+      nrow(scenario_specs_good_to_bad):1,
+    ]
```
maurolepore, line 123: I'm not entirely sure what's happening here but I sense the approach `nrow(x):1` might be fragile. See if `seq_along()` or `seq_len()` might be more robust alternatives for your use case. For example, these all do the same but each is best in some situation:

``` r
seq_along(3:1)
#> [1] 1 2 3
seq_len(3)
#> [1] 1 2 3
1:3
#> [1] 1 2 3
```

<sup>Created on 2021-02-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

--

Nit: I find the if else branches hard to understand at first glance, and mixing extractions with dplyr's `filter()` + `select()` sometimes, and then with `]` doesn't help me much :(

R/plot_2diicharts.R
```diff
@@ -361,6 +362,111 @@ plot_metareport_security_types <- function(data, bars_asset_type_specs =
...
+                                                  "aviation", "automotive", "power",
+                                                  "coal", "oil&gas"
+                                                ),
+                                                "label" = c(
+                                                  "Steel", "Cement", "Shipping",
+                                                  "Aviation", "Automotive", "Power",
+                                                  "Coal", "Oil & Gas"
+                                                )
+                                              ),
+                                              bars_labels_specs = NULL) {
```
maurolepore, line 402: See https://design.tidyverse.org/def-short.html

R/plot_2diicharts.R
```diff
@@ -420,3 +526,24 @@ r2dii_palette_colours <- function() {
...
+    "aviation",   "#00c082",
+    "cement",   "#c1b358",
+    "coal",   "#4e3b37",
+    "oil&gas",   "#181716",
+    "power",   "#a63603",
+    "shipping",   "#574099",
+    "steel",   "#a63d57"
+  )
+  # styler: on
+}
```
maurolepore, line 549: Ups, it looks like we lost the allignment. Is `styler: off/on` not working?

maurolepore (approved):
Thanks! Great to see the gallery of available plots is expanding. What's the roadmap? Do you plan to add many more plots?

MonikaFu:
> Thanks! Great to see the gallery of available plots is expanding. What's the roadmap? Do you plan to add many more plots?

yes, check out issue #10

R/plot_2diicharts.R
cjyetman: @maurolepore like this?
``` r
scenario_specs <- scenario_specs_good_to_bad[rev(seq_len(nrow(scenario_specs_good_to_bad))), ]
```

maurolepore:
Thanks! I see that the number of plots so far is approaching the number of items in https://github.com/2DegreesInvesting/r2dii.ggplot/issues/10 (assuming the checked boxes are outdated). What I wanted to make a guess for how much code you expect to add before we get into the testing stage. It's hard to predict how much faster the project will move without tests and a bit of refactoring -- when considering the mid/long term costs of maintenance. Remember I'm willing to help.

R/plot_2diicharts.R
maurolepore: I just "smell" something but I might be wrong. You are welcome to disregard. Sorry for not being very helpful. I'll be able to give more detailed feedback when what I learn from the code I can capture in some kind of refactoring. But I hear it's early for that.

MonikaFu: this piece of code changes the order of the rows in a data frame. The first row becomes the last one etc. This is the solution that I found on stack overflow. If there is a more elegant way to do it, I'll gladly change it. I also don't really like this piece of code.

R/plot_2diicharts.R
MonikaFu: How could I not have thought about setting it in the code when it is `NULL` 😆

R/plot_2diicharts.R
MonikaFu: it's because I actually created this one manually. Fixed now, thanks for noticing.

MonikaFu:
@maurolepore I would like to close issue #10 before testing. That means 4 more charts. I think that they should be finished by the end of this week. Then we could move to testing.

R/plot_2diicharts.R
maurolepore: Maybe `dplyr::arrange()`?

cjyetman: I guess that would end up as
``` r
scenario_specs <- dplyr::arrange(scenario_specs_good_to_bad, dplyr::desc(dplyr::row_number()))
```

PR #41, opened 2021-02-09
Add map

MonikaFu:
Related to #10 

In this PR I add a map plot to the package:

* data preparation for Equity/Bonds_results_map.rda data
* plotting function

image


R/plot_2diicharts.R
```diff
@@ -712,6 +712,73 @@ plot_metareport_bubble <- function(data,
...
+  p <- ggplot(
+    data,
+    aes(
+      x = .data$long,
+      y = .data$lat,
+      group = .data$group,
+      fill = .data$value
+    )
+  ) +
+    scale_fill_gradientn("", colours = colour_function(9)) +
```
cjyetman, line 761: `scale_fill_gradientn` looks like a spelling error?

MonikaFu: Constant vigilance! But no, it is actually a function that takes `n` colour values as input.

https://rdrr.io/github/UI-Research/urbnthemes/man/scale_fill_gradientn.html

R/data_processing.R
```diff
@@ -355,3 +355,73 @@ prepare_for_metareport_bubble_chart <- function(data_asset_type,
...
+#'   string).
+#' @param year_filter Year to be used for filtering (integer).
+#' @param value_divisor Number by which to divide the values to be plotted
+#'   (integer).
+#' @param allocation_method Allocation method used to calculate the production
+#'   values. If none is specified "portfolio_weight" is used for Bonds and
+#'   "ownership_weight" is used for Equity.
+#'
+#' @return Dataframe with columns (only important columns are listed): long,
+#'   lat, group, value, unit, abbreviation_divisor
```
maurolepore, line 376: Nit: This sentence is missing the final ".".

R/data_processing.R
```diff
@@ -355,3 +355,73 @@ prepare_for_metareport_bubble_chart <- function(data_asset_type,
...
+#' @export
+#'
+#' @examples
+prepare_for_map_chart <- function(data_map_asset_type,
+                                  asset_type,
```
maurolepore, line 381: It's unclear what choices the user has. See https://design.tidyverse.org/def-enum.html

R/data_processing.R
```diff
@@ -355,3 +355,73 @@ prepare_for_metareport_bubble_chart <- function(data_asset_type,
...
+    ungroup() %>%
+    na.omit() %>%
+    mutate(country = iso.expand(.data$ald_location, regex = TRUE)) %>%
+    mutate(country = case_when(
+      .data$country == "(^France)|(^Clipperton Island)" ~ "France",
+      .data$country == "(^China(?!:Hong Kong|:Macao))|(^Paracel Islands)" ~ "China",
+      .data$country == "Norway(?!:Bouvet|:Svalbard|:Jan Mayen)" ~ "Norway",
+      .data$country == "UK(?!r)" ~ "UK",
+      .data$country == "(^Spain)|(^Canary Islands)" ~ "Spain",
+      .data$country == "(^Trinidad)|(^Tobago)" ~ "Trinidad", # assign to Trinidad because it is bigger
```
maurolepore, line 415: Maybe the forcats pacakge has something to express this in a friendlier way? Maybe https://forcats.tidyverse.org/reference/fct_collapse.html

R/data_processing.R
```diff
@@ -355,3 +355,73 @@ prepare_for_metareport_bubble_chart <- function(data_asset_type,
...
+    )) %>%
+    mutate(abbreviation_divisor = case_when(
+      !!value_divisor == 1 ~ "",
+      !!value_divisor == 10^3 ~ "k",
+      !!value_divisor == 10^6 ~ "M",
+      !!value_divisor == 10^9 ~ "B",
+      TRUE ~ as.character(value_divisor)
+    ))
+
+  joined_map <- left_join(map_data("world"), data_map, by = c("region" = "country"))
```
maurolepore, line 426: Do you mean to return invisibly?

``` r
f1 <- function() {
  out <- "I'm invisible"
}

f1()

print(f1())
#> [1] "I'm invisible"



f2 <- function() {
  out <- "I'm visible"
  out
}

f2()
#> [1] "I'm visible"
```

<sup>Created on 2021-02-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/plot_2diicharts.R
```diff
@@ -712,6 +712,73 @@ plot_metareport_bubble <- function(data,
...
 
+#' Create a meta-report map chart
+#'
+#' @param data Dataframe with data processed for the chart. With columns: long,
+#'   lat, group, value (dataframe).
+#' @param plot_title Title of the plot; (character string).
+#' @param legend_title Title of the legend describing value units; (character
+#'   string).
+#' @param sector Sector to be used for the map colouring. If none is specified,
+#'   default dark blue palette is used (character string)
```
maurolepore, line 723: Nit: Missing "." at the end.

R/plot_2diicharts.R
```diff
@@ -712,6 +712,73 @@ plot_metareport_bubble <- function(data,
...
+    dark_colour <- colours %>%
+      filter(.data$label == "dark_blue") %>%
+      pull(.data$colour_hex)
+  } else {
+    sec_colours <- r2dii_sector_colours()
+
+    dark_colour <- sec_colours %>%
+      filter(.data$label == tolower(!!sector)) %>%
+      pull(.data$colour_hex)
+  }
```
maurolepore, line 745: Maybe extract into `choose_colours(sector)` or similar?

R/plot_2diicharts.R
maurolepore: Which package is the function coming from? Do we need to depends on a new package?

(Maybe you need to call `use_package("pkgname")` or `use_dev_package("pkgname")`?)

README.Rmd
```diff
@@ -224,3 +224,32 @@ p <- plot_metareport_bubble(data_bubble,
                             )
 p
 ```
+These is a basic example of using `plot_metareport_map()` given that
+you have the `Equity_results_map.rda` or `Bonds_results_map.rda`
+data set from PACTA analysis results loaded in your environment:
+
+```
```
maurolepore, line 231: This code won't run. Is this intentional?

maurolepore (approved):

R/data_processing.R
cjyetman: or [countrycode](https://github.com/vincentarelbundock/countrycode)

R/plot_2diicharts.R
MonikaFu: @maurolepore it is actually a part of `ggplot2` which we import whole. If it wasn't a part of packages imported already the check would throw an error, wouldn't they?

MonikaFu: To be precise - not an error but a note. But still - I could only run into issues you both are mentioning above if I didn't run the 'check', which I always do before submitting a PR.

Screen Shot 2021-02-10 at 2 17 36 PM

README.Rmd
MonikaFu: yes, I have no example data set to run it with for now.

R/plot_2diicharts.R
MonikaFu: does it make sense if this is the only place in which I perform this operation?

R/data_processing.R
MonikaFu: not necessarily

R/data_processing.R
MonikaFu: going to use countrycode, it's better indeed.

R/data_processing.R
MonikaFu: I see the point and I need to do it in many places so I'm going to make an issue for it.

cjyetman:
One suggestion with `countrycode`.... 2dii frequently uses "XK" as an iso2c even though technically it's not a true iso2c. You can make `countrycode` work with that by using the `custom_match` argument...

``` r
library(countrycode)
country_isos <- c("US", "DE", "XK")

countrycode(country_isos, "iso2c", "country.name")
#> Warning in countrycode(country_isos, "iso2c", "country.name"): Some values were not matched unambiguously: XK
#> [1] "United States" "Germany"       NA

countrycode(country_isos, origin = "iso2c", destination = "country.name", 
            custom_match = c("XK" = "Kosovo"))
#> [1] "United States" "Germany"       "Kosovo"
```

R/plot_2diicharts.R
maurolepore: It makes sense to me but not everyone.It's your call. I like when the body of a function if mostly doing what the name of the function promises to do. In my opinion the details about how to choose a color bury the main goal of the function -- which by its name I infer is to create a plot.

I opened a discussion that is somewhat related. Here I show a high level function that I tried to keep at a consistent high-level of detail: https://github.com/2DegreesInvesting/practices/discussions/10#discussioncomment-341558

MonikaFu:
@cjyetman - so when I use `countrycode` instead of `maps` for the iso codes I get a very different map:

image

Any opinions on that? I'm asking because I saw that you are a maintainer of the package. I'm going to investigate further to see where the differences exactly are.

MonikaFu:
Ok, I see what is happening I think, some countries in `countrycode` are not compatible with ggplot `map_data("world")`. Going to fix this when back to work.

cjyetman:
So, `ggplot2` apparently uses a list of country names and regions with specific, but seemingly arbitrary spellings as its identifier for map data (which sucks, they obviously should use a standardized country code). In that type of situation (you want to convert a vector of country names or codes into some specific set of spellings), I usually suggest building a custom dictionary and using that with `countrycode` (which is a supported feature/process).

I take the full map data from ggplot, convert its country/region names to iso2c, remove duplicate rows, remove rows where an iso2c could not be determined, remove rows where the iso2c is duplicated (more on that later), and convert it to a tibble for easy inspecting/printing.

``` r
library(dplyr)
library(countrycode)
library(ggplot2)

ggplot_countries <- 
  ggplot2::map_data("world") %>% 
  mutate(iso2c = countrycode(region, "country.name", "iso2c", 
                             custom_match = c("Kosovo" = "XK"),
                             warn = FALSE)) %>% 
  select(iso2c, ggplot = region) %>%
  distinct() %>% 
  filter(!is.na(iso2c)) %>% 
  filter(!duplicated(iso2c)) %>% 
  as_tibble()
```

that custom dictionary can then be used like this...
``` r
countrycode(data_map_eq$ald_location, 
            origin = "iso2c", destination = "ggplot", 
            custom_dict = ggplot_countries)
```

Removing the duplicated iso2c codes is necessary because `ggplot2` has separate identifiers for islands that share a unified iso2c code. Without the `filter(!duplicated(iso2c))` line, you'll end up with the following. I took the easy route and just chose the first one of each pair, but you/one may have a different preference on how to deal with that.
``` r
ggplot_countries[duplicated(ggplot_countries$iso2c, fromLast = T) | duplicated(ggplot_countries$iso2c, fromLast = F), ]
# # A tibble: 6 x 2
#   iso2c ggplot                
#   <chr> <chr>                 
# 1 KN    Nevis                 
# 2 KN    Saint Kitts           
# 3 GS    South Sandwich Islands
# 4 GS    South Georgia         
# 5 TT    Trinidad              
# 6 TT    Tobago     

countrycode(c("KN", "GS", "TT"), "iso2c", "country.name")
#> [1] "St. Kitts & Nevis"                     
#> [2] "South Georgia & South Sandwich Islands"
#> [3] "Trinidad & Tobago"
```

and lastly, my test PACTA outputs had two iso2c codes that don't seem to have any appropriate identifier in the `ggplot2` data...
``` r
countrycode(c("HK", "VI"), "iso2c", "country.name")
#> [1] "Hong Kong SAR China" "U.S. Virgin Islands"
```

After I wrote that, I found that `ggplot2` has an appropriate two part identifier for `VI` with `region == "Virgin Islands" & subregion == "US"`. Not sure how "US" is a subregion... hehehe... but anyway, you could append a line to the custom ggplot country lookup table for that, but I guess technically you'd have to also add the subregion code when modifying your data so that you don't plot its data on the British Virgin Islands also.

MonikaFu:
@cjyetman thanks for your extensive answer! What you suggest sounds very reasonable, I will try to implement it.

cjyetman:
I just looked into the map think further... or from a different perspective...

An alternate (maybe better?) solution would be to use `maps::iso.alpha()` from the `maps` package, which is where the data from `ggplot2::map_data` comes from...
``` r
world_map <- ggplot2::map_data(map = "world")
world_map$iso3c <- maps::iso.alpha(world_map$region, n = 3)

joined_map <- left_join(world_map, data_map, by = c("iso3c" = "country"))
```

MonikaFu:
@cjyetman thanks for the input. I'm working now on finding the best solution, looking into both options.

MonikaFu:
In the end I decided to use `iso.alpha` function from `maps` package as suggested by @cjyetman and more in line with what was used initially. It is more compatible with `map_data` from `ggplot2`. Instead of expanding the ISO codes that we use in PACTA data, I add a `iso2c` code to world map data and join the world map with PACTA output using the 2 letter ISO code.

maurolepore (approved):
LGTM

Issue #26, opened 2021-01-22
Add tests


maurolepore:
User story: **As a** contributor to this package **I would like to have** unit tests **so that I can** know when my contribution breaks existing behaviour.

[Example](https://github.com/forestgeo/fgeo.plot/blob/master/tests/testthat/test-plot_dbh_bubles_by_quadrat.R)

PR #50, opened 2021-02-12
Add tests

maurolepore:
Relates to #26 

This PR adds a few tests. I pause here because it can feel overwhelming. We can discuss what I did here and plan the next steps.

In general, the changes here do not change existing behavior (when they do I flag it as a comment below). Where I see a problem I write a "charactirization test" that exposes the problem and I add `# FIXME` or `# TODO`. You may run a gobal serach to quickly find all those places.


DESCRIPTION
```diff
@@ -21,6 +21,11 @@ Imports:
     magrittr,
     r2dii.data,
     scales
+Suggests: 
+    r2dii.analysis,
+    r2dii.match,
```
maurolepore, line 26: The helpfile of `get_example_data()` "promises" to return a data set that is like a dataset in other r2dii packages. To test this promise I'm suggesting r2dii packages so we can use them in tests and keep R CMD check happy. This will install new packages if they are not already installed and therefore changes previous behaviour.

R/get_example_data.R
```diff
@@ -1846,6 +1847,6 @@ get_example_data <- function() {
     )), skip = 1L
   ), class = "col_spec"))
 
-
+  data$year <- as.integer(data$year)
```
maurolepore, line 1850: A test showed that the output of `get_example_data()` was not structured exactly as promised in the documentation. So I changed the source code to match the expectation. This change changes existing behaviour.

README.Rmd
```diff
@@ -50,9 +50,13 @@ example_data <- get_example_data()
 example_data <- process_input_data(example_data)
 
 data_trajectory <- prepare_for_trajectory_chart(
-  data = example_data, sector_filter = "power", technology_filter = "oilcap",
```
maurolepore, line 53: The argument name is now `data_processed` not `data` (it works because R does partial matching but it's not a good idea). Instead of fixing the name I removed it -- this is fine for the first argument and sometimes the second too.

maurolepore:
Use the tidier version #51

PR #51, opened 2021-02-13
Add tests

maurolepore:
Relates to #26 
(This is a tidier version of  #50.)

This PR adds a few tests. I pause here because it can feel overwhelming. We can discuss what I did here and plan the next steps.

In general, the changes here do not change existing behavior (when they do I flag it as a comment below). Where I see a problem I write a test that exposes it and I add a comment starting with `# FIXME` or `# TODO`.


R/get_example_data.R
```diff
@@ -1846,6 +1847,6 @@ get_example_data <- function() {
     )), skip = 1L
   ), class = "col_spec"))
 
-
+  data$year <- as.integer(data$year)
```
maurolepore, line 1850: This changes existing behaviour but I think it's for the better and I don't think it will break anything.

DESCRIPTION
```diff
@@ -21,6 +21,11 @@ Imports:
     magrittr,
     r2dii.data,
     scales
+Suggests: 
+    r2dii.analysis,
+    r2dii.match,
```
maurolepore, line 26: This changes existing behavour by installing this packages if they are not already installed. This should brake nothing and allows us to write integration tests which keeping R CMD check happy.

maurolepore (commented):
These changes affect existing behaviour but they are safe.

MonikaFu (approved):
Your changes LGTM Mauro. Following your methodology I developed a set of tests for tech mix chart which still need to be reviewed before we merge this branch in.

DESCRIPTION
```diff
@@ -27,10 +27,10 @@ Imports:
     tidyr,
     stats,
     grDevices,
-    maps
-Suggests: 
-    r2dii.analysis,
+    maps,
     r2dii.match,
+    r2dii.analysis
+Suggests:
```
MonikaFu, line 33: @maurolepore - the checks threw an error when these packages were given as suggested only so I added them using `usethis::use_package()`. Was that to be expected?

MonikaFu:
@maurolepore - I still get errors in the checks and don't know why :(

DESCRIPTION
maurolepore: Can you show me the error?

maurolepore: I would expect "Suggests" to suffice if the packages are used only in examples and tests. If they are used in main functions, then they need to be in "Imports".

MonikaFu: you can see them in here below - I meant the R-CMD-checks.

maurolepore:
> @maurolepore - I still get errors in the checks and don't know why :(

Which error do you mean? I assume you get the errors I see in the checks on gh-actions. There I see one error that suggests that the output of `get_example_data()` changed in some way, so the reference snapshot may need to be updated by running this:

```r
snapshot_accept('get_example_data')
```

When looking at the errors on gh-actions, I noticed some errors appeared in only some platforms. Before I look at that too deeply, I decided to rerun all checks. Those kinds of inconsistencies suggest problems in the servers instead of our code.

Let's see what the new results show.

maurolepore:
For the record, on gh-actions I see these errors.


### macos

```
── Install failure ─────────────────────────────────────────────────────────────

* installing *source* package ‘r2dii.ggplot’ ...
** using staged installation
** R
** inst
** byte-compile and prepare package for lazy loading
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called ‘ggsignif’
Calls: <Anonymous> ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
ERROR: lazy loading failed for package ‘r2dii.ggplot’
* removing ‘/Users/runner/work/r2dii.ggplot/r2dii.ggplot/check/r2dii.ggplot.Rcheck/r2dii.ggplot’

1 error ✖ | 0 warnings ✔ | 0 notes ✔
Error: Process completed with exit code 1.
```


### ubuntu

```
Error: Error: R CMD check found ERRORs
...

Run `snapshot_accept('get_example_data')` if this is a deliberate change

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 31 ]
Error: Error: Test failures
```

...


...

DESCRIPTION
MonikaFu: > I would expect "Suggests" to suffice if the packages are used only in examples and tests. If they are used in main functions, then they need to be in "Imports".

You are right. I was getting errors locally because I did not have those packages installed at all. Do you suggest reverting the last commit (b6031b8) or just make a new commit moving those packages to 'Suggested' again?

MonikaFu:
I see exactly the same errors.

MonikaFu:
When I run `test that::snapshot_accept('get_example_data')` I get a message back:
`No snapshots to update`

So maybe indeed the checks are failing because of some server issues.

maurolepore:
Yeah, snapshot tests can be fragile. Locally I see all tests and checks pass cleanly. Do you see the same? If so, I wouldn't worry too much just now. I would merge and move on.

maurolepore:
You may also add `testthat::skip_on_ci()` at the top of the problematic test. We may want to remove it later but by then maybe the servers will be working as we expect.

MonikaFu:
@maurolepore yes - locally all the checks pass cleanly for me. I am going to proceed and merge this PR assuming that what we see are server issues.

PR #56, opened 2021-02-25
Add sda example dataset

### LINKED ISSUE #55: Add example sda dataset
For the emission timeline plot an sda dataset is required. Add this as a dataset in the package.

MonikaFu:
Closes #55 

In this PR I add an example dataset with sda targets to the package. I follow closely the checklist from #2 and the practices I observed in `r2dii.data` repo.


tests/testthat/test-sda_target.R
```diff
@@ -0,0 +1,24 @@
+test_that("outputs a data.frame", {
+  expect_s3_class(sda_target, "data.frame")
+})
+
+test_that("has 208 rows and 4 columns", {
+  expect_equal(nrow(sda_target), 208)
+  expect_equal(ncol(sda_target), 4)
+})
```
maurolepore, line 8: If you already have a snapshot, this is redundant. See that the snapshot already includes this information:

```r
      # A tibble: 208 x 4
         sector      year emission_factor_metric emission_factor_value
         <chr>      <dbl> <chr>                                  <dbl>
```

When two tests test the same thing, a single problem will cause two tests to fail when one would be enough and easier to maintain.

data-raw/sda_target.R
```diff
@@ -0,0 +1,3 @@
+# Source: @MonikaFu https://github.com/2DegreesInvesting/r2dii.ggplot/issues/55
+sda_target <- readr::read_csv(here::here("data-raw", "sda_target.csv"))
+usethis::use_data(sda_target, overwrite = TRUE)
```
maurolepore, line 3: I think this is all you need about `sda_data`. Adding a dictionary seems like an overkill. We do add a dictionary in r2dii.data but only because the main goal of that package is precisely to host datasets -- many of them. But the goal of this package is very different. 

 @jdhoffa, did you mean to argue that this package needs the same infrastructure we use in r2dii.data or did you only mean to show off :smile: ?

R/sda_target.R
```diff
@@ -0,0 +1,17 @@
+#' Time series of emission factors
+#'
+#' An example data set containing emission factors for different metric types:
+#' projected, benchmark and scenarios and different sectors.
+#'
+#' @format A data frame with 208 rows and 4 variables:
```
maurolepore, line 6: This should be automatically generated by roxygen. No need to hard code it.

R/sda_target.R
```diff
@@ -0,0 +1,17 @@
...
+#' \describe{
+#'   \item{sector}{industry sector (as defined in PACTA)}
+#'   \item{year}{year for which emissions are reported}
+#'   \item{emission_factor_metric}{metric type: 'projected' for the test
+#'   portfolio 'corporate_economy' for the benchmark, 'adjusted_scenario_demo'
+#'   for a scenario}
+#'   \item{emission_factor_value}{emissions value, a double, units differ
+#'   depending on the sector}
+#' }
```
maurolepore, line 15: You may use rmardkown syntax. See https://usethis.r-lib.org/reference/use_roxygen_md.html

With rmarkdown syntax, and a few tweaks, this documentation becomes:

```r
#' Time series of emission factors
#'
#' An example data set containing emission factors for different metric types:
#' projected, benchmark and scenarios and different sectors.
#' 
#' Column definitions:
#' * `sector`: Industry sector (as defined in PACTA).
#' * `year`: Year which emissions are reported.
#' ...
#'
#' @source <https://github.com/2DegreesInvesting/r2dii.ggplot/issues/55>.
#' @examples
#' sda_targets
"sda_target"
```

maurolepore (approved):
LGTM but I would remove all that relates to the data dictionary. Please confirm with @jdhoffa.

BTW excellent choice of tests!

R/sda_target.R
MonikaFu: how could I trigger creating it then?

MonikaFu:
Thank you @maurolepore for your review. I'm still curious to hear the opinion of @jdhoffa about the data dictionary. There will be a couple of more example datasets in this package but not many more. Also now that I think about it - maybe at some point it will be a better idea to source the example datasets from `r2dii.data` if that is possible?

maurolepore:
Moving example datasets to r2dii.data makes sense to me too. You are welcome to open a PR in r2dii.data and maintain it until just before this package goes to CRAN. Only at that stable point I would merge the PR into r2dii.data and submit the updated r2dii.data (with this example data) to CRAN.

R/sda_target.R
maurolepore: When you `devtools::document()` your package (often indirectly via `devtools::check()`) roxygen reads your files in R/ and creates a corresponding .Rd file inside man/. There is nothing manual you need to do.

For an example consider `r2dii.data::ald_demo`:

* [R/ald_demo.R](https://github.com/2DegreesInvesting/r2dii.data/blob/master/R/ald_demo.R)
* [man/ald_demo.Rd](https://github.com/2DegreesInvesting/r2dii.data/blob/master/man/ald_demo.Rd)
* [helpfile (ald_demo.html)](https://2degreesinvesting.github.io/r2dii.data/reference/ald_demo.html)

data-raw/sda_target.R
jdhoffa: ... i was showing off :-(

jdhoffa:
Absolutely agree that it could live in `r2dii.data` once `r2dii.ggplot` is on CRAN. Further, I agree that the data dictionary is a overkill at this stage. Apologies for wasting your time, I honestly just copied and pasted the workflow I use to add datasets to `r2dii.data` without giving it a lot of thought.

R/sda_target.R
MonikaFu: right. But I did not create an .Rd file manually.. Or is that not what you mean? I basically followed what I've seen being done in `r2dii.data`.

MonikaFu:
> Absolutely agree that it could live in `r2dii.data` once `r2dii.ggplot` is on CRAN. Further, I agree that the data dictionary is a overkill at this stage. Apologies for wasting your time, I honestly just copied and pasted the workflow I use to add datasets to `r2dii.data` without giving it a lot of thought.

And I was following your advice like a good student @jdhoffa 😆 alright, I'm going to remove the data dictionary then since it will make the maintenance of the package easier.

R/sda_target.R
maurolepore: What I mean but didn't say explicitely is that your documentation in R/ should not include a `#' @format` section. From R/sda_target.R You can delete this:

```
#' @format A data frame with 208 rows and 4 variables:
```

 That is desirable because it's too easy to forget to update the number of rows and columns when the size of `sda_target` changes.

MonikaFu: Ah, I see. thanks.

PR #57, opened 2021-02-25
Suggest problematic 'ggsignif' package from GitHub

maurolepore:
ggsignif is a dependency of ggpubr. Usually we do not need to install a dependency of a dependency but here I do it as a hack.

I don't know why, but GitHub actions is failing to install ggsignif from CRAN. To get around this problem, this PR installs ggsignif from GitHub instead of CRAN. This won't be acceptable when you submit to CRAN, but by then I hope to have pruned as many dependencies as possible, maybe including ggpubr -- which seems quite heavy.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,8 @@ Imports:
 Suggests: 
     r2dii.analysis,
     r2dii.match,
-    testthat (>= 3.0.0)
+    testthat (>= 3.0.0),
+    ggsignif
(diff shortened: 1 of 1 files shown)


maurolepore:
Notice CI no longer fails due to ggsignif. The problem I still see is irrelevant to this PR and is addressed at #58.

MonikaFu (approved):
I confirm - the errors are not due to `ggsignif`. LGTM

PR #58, opened 2021-02-25
Store a more robust snapshot

maurolepore:
This PR fixes a problem I saw on CI. A snapshot was failing because the print method of tibble was slightly different locally versus CI. Tibble likely includes a way to deal with this more elegantly, but instead here I use the simpler approach of converting the tibble to a vanilla dataframe -- which prints the same everywhere.

(As of this writing I see some checks failing on macos and windows but for a reason that is irrelevant to this PR. That problem is fixed at #57.)

### DIFF
--- a/tests/testthat/_snaps/get_example_data.md
+++ b/tests/testthat/_snaps/get_example_data.md
@@ -1,20 +1,20 @@
 # outputs the expected snapshot
 
     Code
-      get_example_data()
+      some_example_data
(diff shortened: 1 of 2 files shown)


MonikaFu (approved):
LGTM. This problem should also be solved I hope once I convert this dataset to a dataset in a package and won't be creating it using `get_example_data()`.

PR #59, opened 2021-03-09
Add emissions time line

### LINKED ISSUE #7: Add emissions projection chart using SDA data
A new graph to be added to the library

MonikaFu:
Closes #7 

In this PR I add:
* data processing function for sda banks' output and timeline plot
* plotting function for timeline

image

FYI @jdhoffa


tests/testthat/test-plot_timelines.R
```diff
@@ -0,0 +1,61 @@
+test_that("prints output ggplot object without error", {
+  data_sda_cement <- prepare_for_timeline(sda_target,
+                                      sector_filter = "cement",
+                                      year_start = 2020,
+                                      year_end = 2026,
+                                      column_line_names = "emission_factor_metric",
+                                      value_to_plot = "emission_factor_value")
+
+  plot <- plot_timelines(data_sda_cement)
+  expect_error(print(plot),NA)
```
MonikaFu, line 10: @maurolepore - the result of this function is a .pdf plot that is saved in the testing directory which I feel is undesirable. Any ideas how to avoid this? I think that checking whether it is possible to `print` the plot is a useful test.

MonikaFu (commented):
@maurolepore I am considering using [vdiffr package](https://github.com/r-lib/vdiffr) for testing the visual output. It is integrated with `testthat` apparently. Any thoughts on that? Good idea or not desirable to introduce another dependency?

tests/testthat/test-plot_timelines.R
maurolepore: I think testing visuals is a good idea. I haven't done it myself so you are about to become our local expert :-).
I think it's fine to depend on vdiffr -- I trust the maintainer and you only need it in suggests (`usethis::use_package("vdiffr", "suggests")`). But vdiffr predates testthat e3, which implements [snapshot testing](https://testthat.r-lib.org/articles/snapshotting.html).


 is maintained by the same group that developed testthat.

maurolepore: > @maurolepore - the result of this function is a .pdf plot that is saved in the testing directory which I feel is undesirable. Any ideas how to avoid this? I think that checking whether it is possible to print the plot is a useful test.

Maybe print to a temporary file instead.

``` r
library(ggplot2)

dataset <- datasets::BOD
tmp <- tempfile(fileext = ".pdf")

ggplot(dataset) + geom_point(aes(Time, demand))
```



``` r
ggsave(tmp)
#> Saving 7 x 5 in image
file.exists(tmp)
#> [1] TRUE

on.exit(unlink(tmp), add = TRUE)
file.exists(tmp)
#> [1] FALSE
```

<sup>Created on 2021-03-09 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

tests/testthat/test-plot_timelines.R
```diff
@@ -0,0 +1,61 @@
+test_that("prints output ggplot object without error", {
+  data_sda_cement <- prepare_for_timeline(sda_target,
+                                      sector_filter = "cement",
+                                      year_start = 2020,
+                                      year_end = 2026,
+                                      column_line_names = "emission_factor_metric",
+                                      value_to_plot = "emission_factor_value")
```
maurolepore, line 7: style?

R/plot_timelines.R
```diff
@@ -0,0 +1,84 @@
...
+#' @param x_title,y_title x- and y-axis title.
+#'
+#' @return An object of class "ggplot".
+#' @export
+#'
+#' @examples
+#' data_sda_cement <- prepare_for_timeline(sda_target,
+#'                                      sector_filter = "cement",
+#'                                      year_start = 2020,
+#'                                       year_end = 2026,
```
maurolepore, line 17: What happens when the data lacks this years? Maybe instead default to the minimum and maximum value in some `year` colum?

R/plot_timelines.R
```diff
@@ -0,0 +1,84 @@
...
+#'   (optionally) colour names from the r2dii_colours palette.
+#' @param plot_title Title of the plot.
+#' @param x_title,y_title x- and y-axis title.
+#'
+#' @return An object of class "ggplot".
+#' @export
+#'
+#' @examples
+#' data_sda_cement <- prepare_for_timeline(sda_target,
+#'                                      sector_filter = "cement",
```
maurolepore, line 15: Is "cement" common enough to be a default? What other options we have? Maybe use `c("cement", "<other>", "<other2>", ...)` instead?

R/prepare_for_timeline.R
```diff
@@ -0,0 +1,64 @@
...
+prepare_for_timeline <- function(sda_target_data,
+                                 sector_filter = c("automotive",
+                                                   "aviation",
+                                                   "cement",
+                                                   "oil and gas",
+                                                   "shipping",
+                                                   "steel",
+                                                   "power"),
+                                 year_start = 0,
+                                 year_end = Inf,
```
maurolepore, line 33: Is it possible to use defaults consistently across all funcitons in the package? That way, once a user learns one function they can predict how to use all other functions.

tests/testthat/test-plot_timelines.R
```diff
@@ -0,0 +1,61 @@
...
+test_that("with bad 'lines_specs' errors gracefully", {
+  data_sda_cement <- prepare_for_timeline(sda_target,
+                                      sector_filter = "cement",
+                                      year_start = 2020,
+                                      year_end = 2026,
+                                      column_line_names = "emission_factor_metric",
+                                      value_to_plot = "emission_factor_value")
+
+  bad <- "bad"
+  graceful_message <- "'line_specs' needs to be a dataframe with columns"
```
maurolepore, line 22: Great we got to this point :-)

Now you may want to read https://style.tidyverse.org/error-messages.html

tests/testthat/test-plot_timelines.R
```diff
@@ -0,0 +1,61 @@
...
+test_that("with bad colour names in 'lines_specs' errors gracefully", {
+  data_sda_cement <- prepare_for_timeline(sda_target,
+                                      sector_filter = "cement",
+                                      year_start = 2020,
+                                      year_end = 2026,
+                                      column_line_names = "emission_factor_metric",
+                                      value_to_plot = "emission_factor_value")
+
+  bad_lines_specs <- data.frame(
+  "line_name" = c("projected", "corporate_economy", "target_demo", "adjusted_scenario_demo"),
```
maurolepore, line 53: Maybe `tolower(label)` is better than duplicating it.

tests/testthat/test-prepare_for_timeline.R
```diff
@@ -0,0 +1,85 @@
...
+                              year_end = 2026,
+                              column_line_names = "emission_factor_metric",
+                              value_to_plot = "emission_factor_value")
+
+  expect_s3_class(out, "data.frame")
+})
+
+test_that("with bad `sector_filter` errors gracefully", {
+  bad <- "bad"
+  graceful_message <- "'arg' should be one of "
```
maurolepore, line 14: In messages to the user, it may be useful to format different kind of objects differently. For example, the tidyverse uses "value" and `column` -- not "column" and not `value`.

For inspiration you may see `usethis::ui_*()` functions

https://usethis.r-lib.org/reference/ui.html

maurolepore (approved):
LGTM :-)

tests/testthat/test-plot_timelines.R
MonikaFu: I've looked into snapshot testing for plots and it indeed seems to be possible. However, I did not manage to get it to work. I was trying to follow this website: https://testthat.r-lib.org/reference/expect_snapshot_file.html but it wasn't clear enough for me how to save the snapshot etc. If you could help me here @maurolepore it would be great.

R/plot_timelines.R
MonikaFu: it is a category that is in the test dataset. I don't expect that to change so why should I give more values? I don't fully understand but I also don't know exactly how the examples work.

R/plot_timelines.R
MonikaFu: The test data contains these years. In general they are used for filtering with `<=` and `>=` so as long as the data has some years in between `year_start` and `year_end` it works.

R/prepare_for_timeline.R
MonikaFu: can you add it as an issue?

tests/testthat/test-plot_timelines.R
MonikaFu: I run styler over these..

tests/testthat/test-plot_timelines.R
MonikaFu: I don't understand what you mean exactly.

tests/testthat/test-prepare_for_timeline.R
MonikaFu: can you say what would you change in this particular example? I don't follow your line of thought here..

maurolepore: In this particular example 'arg' (surrounded by quotes) becomes `arg` (surrounded by back-tics). 

``` r
library(usethis)

ui_code("an_argument")
#> `an_argument`
ui_value("a_value")
#> 'a_value'
```

<sup>Created on 2021-03-29 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

tests/testthat/test-plot_timelines.R
maurolepore: Notice how both produce the same result, but the "Bad" version is harder to maintain because it duplicates code:

``` r
# Good
tibble::tibble(label = "Projected", line_name = tolower(label))
#> # A tibble: 1 x 2
#>   label     line_name
#>   <chr>     <chr>    
#> 1 Projected projected

# Bad
tibble::tibble(label = "Projected", line_name = "projected")
#> # A tibble: 1 x 2
#>   label     line_name
#>   <chr>     <chr>    
#> 1 Projected projected
```

<sup>Created on 2021-03-29 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

R/plot_timelines.R
maurolepore: My mistake. I seem to have confused an example with a function definition.

R/plot_timelines.R
maurolepore: The function definition now looks correct

```r 
# ...
                                 year_start = 0,
                                 year_end = Inf,
# ...
```

Maybe it was always correct and I confused an example with a function definition.

jdhoffa:
Quikc comment: the timeline should go to 2050 in the SDA case (not to 2025). 
SDA is a convergence target, where the convergence date is 2050 (but note the ALD and portfolio data will only go to 2025, but the target and scenario data will go to 2050. 
@MonikaFu apologies if this has already been changed in a more recent commit, I have not yet checked this out locally (it's been a while since I've been coding and need to update EVERYTHING lol)

jdhoffa:
Ok, it looks like this is fixed easily as an argument to `prepare_for_timeline()`. The example still plots only to 2025, so maybe just update the argument `year_end` to `2050`?

``` r
library(r2dii.ggplot)

data_sda_cement <- prepare_for_timeline(sda_target,
                                        sector_filter = "cement",
                                        year_start = 2020,
                                        year_end = 2050,
                                        column_line_names = "emission_factor_metric",
                                        value_to_plot = "emission_factor_value"
)

lines_specs <- data.frame(
  "line_name" = c("projected", "corporate_economy", "target_demo", "adjusted_scenario_demo"),
  "label" = c("Projected", "Corporate Economy", "Target Demo", "Adjusted Scenario Demo"),
  "r2dii_colour_name" = c("dark_blue", "green", "grey", "orange")
)

plot_timeline(data_sda_cement,
              lines_specs = lines_specs,
              plot_title = "Emission intensity trend for Cement.",
              x_title = "Year",
              y_title = "Tons of CO2 per ton of Cement"
)
```



<sup>Created on 2021-04-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

README.Rmd
```diff
@@ -161,6 +161,32 @@ plot_techmix_custom_col <- plot_techmix(data_techmix_power, "Technology mix for
...
 ```
 
+* `prepare_for_timeline()` `r pull_title("prepare_for_timeline")`.
+* `plot_timeline()` `r pull_title("plot_timeline")`.
+
+```{r}
+data_sda_cement <- prepare_for_timeline(sda_target,
+                                      sector_filter = "cement",
+                                      year_start = 2020,
+                                      year_end = 2050,
```
MonikaFu, line 171: @jdhoffa - in the recent README version the `year_end` is 2050.

jdhoffa: Ok cool, then I would just change it in the @examples

R/plot_timeline.R
```diff
@@ -0,0 +1,143 @@
...
+#' @param x_title,y_title x- and y-axis title.
+#'
+#' @return An object of class "ggplot".
+#' @export
+#'
+#' @examples
+#' data_sda_cement <- prepare_for_timeline(sda_target,
+#'   sector_filter = "cement",
+#'   year_start = 2020,
+#'   year_end = 2026,
```
jdhoffa, line 18: @MonikaFu

README.Rmd
MonikaFu: OMG, so many places for updating 🙈 will do!

tests/testthat/test-plot_timelines.R
MonikaFu: I see your point however the `line_name` is not exactly `tolower(label)`. The `label` is quite arbitrary and `line_name` needs to match the data. So I leave it as is.

tests/testthat/test-prepare_for_timeline.R
MonikaFu: This error message is actually automatically generated when you specify potential values for an argument and then run the function `match.arg()` on what was supplied by the user. So it would be quite difficult to change to the convention you propose. And might be confusing I'd say.

MonikaFu:
@maurolepore - I know that you approved this PR already but since I've made quite some changes since then and addressed all your comments (I think) - could you re-confirm the approval? 

Also - some checks are failing but I think that it's not due to a bug in a code but some environment issues. What do you think? Should I double check? In one of the tests I compare if two arrays are identical and I think that this is the test that causes issues. Maybe you can have an idea how to improve my method ? Will indicate in the code as well .

R/plot_timeline.R
```diff
@@ -0,0 +1,142 @@
...
+    msg <- paste0(
+      "'line_specs' must have columns 'line_name' and 'label'.\n",
+      paste0("* Yours has columns: ", toString(names(lines_specs)), ".\n"),
+      "* Optionally you could add a column 'r2dii_colour_name'."
+    )
+    stop(msg)
+  }
+
+  if (!identical(sort(unique(lines_specs$line_name)), sort(unique(data$line_name)))) {
+    msg <- paste0(
```
MonikaFu, line 63: @maurolepore - here is where I compare the two arrays to see if `line_specs` contains specs for all the data in `data`. Maybe my test is too strict and that is why some of the checks are failing? Or there is a better way to do it?

README.Rmd
maurolepore: To facilitate maintenance you may keep README bare bones. See for example [dplyr's README](https://dplyr.tidyverse.org/): Although dplyr includes a ton of useful functions, README only shows a handful. If users want more details they can see the helpfiles (also available via the Reference section of the website -- if you add one).

R/plot_timeline.R
maurolepore: I see no reason why these two would be identical with R-release (4.0) but  not with R-oldrel (3.6):

```r
sort(unique(lines_specs$line_name)), sort(unique(data$line_name))
```

Luckily the problem shows up both on Windows and Ubuntu 16.04 -- so I could try to reproduce the issue in a docker container running Ubuntu 16.04 and R 3.6.

MonikaFu: Me neither.. Would be great if you would be able to replicate this issue!

maurolepore: I'll be busy with my performance review for about 1h. If you are curious you may try it yourself on rstudio.cloud. 

* Create a new project from your github repo

image


* Select R 3.6.3

image

R/plot_timeline.R
```diff
@@ -0,0 +1,142 @@
...
+#'                                      year_start = 2020,
+#'                                      year_end = 2050,
+#'                                      column_line_names = "emission_factor_metric",
+#'                                      value_to_plot = "emission_factor_value",
+#'                                      extrapolate_missing_values = TRUE)
+#'
+#'lines_specs <- data.frame(
+#'  "line_name" = c("projected", "corporate_economy", "target_demo", "adjusted_scenario_demo"),
+#'  "label" = c("Projected", "Corporate Economy", "Target Demo", "Adjusted Scenario Demo"),
+#'  "r2dii_colour_name" = c("dark_blue", "green","grey","orange")
```
maurolepore, line 26: Here you need stringsAsFactors = FALSE or use `tibble::tibble()`.

maurolepore (changes requested):
I found the bug. `data.frame()` default of `stringsAsFactor` changed from `TRUE` to false in R 4.0. This is the most important breaking change of 4.0 so I should have thought of it!

If appropriate, use `data.frame()` with `stringsAsFactors = FALSE` or use `tibble::tibble()` instead. The error comes from only two instances buy consider that changet for every instance of `data.frame()`.

R/plot_timeline.R
maurolepore:
``` r
# R < 4.0
default <- TRUE
data <- data.frame(x = "a", stringsAsFactors = default)

identical("a", data$x)
#> [1] FALSE

class("a")
#> [1] "character"
class(data$x)
#> [1] "factor"

# R >= 4.0
default <- FALSE
data <- data.frame(x = "a", stringsAsFactors = default)

identical("a", data$x)
#> [1] TRUE

class("a")
#> [1] "character"
class(data$x)
#> [1] "character"
```

<sup>Created on 2021-04-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

R/plot_timeline.R
```diff
@@ -0,0 +1,142 @@
...
+  if (is.null(lines_specs)) {
+    lines_specs <- data.frame(
+      "line_name" = unique(data$line_name),
+      "label" = unique(data$line_name)
+    )
+  } else if (typeof(lines_specs) != "list") {
+    msg <- paste0("'line_specs' must be a dataframe.\n",
+                  paste0("* You've supplied a ", typeof(lines_specs), "."))
+    stop(msg)
+  }
```
maurolepore, line 49:
``` r
# If using `stop()`, then `paste0()` is unnecessary
f <- function() stop("Some letters are: ", head(letters), ".")
f()
#> Error in f(): Some letters are: abcdef.

# The tidyverse style guide recommends `call. = FALSE` because information about
# where the error happened is easy to recover with `traceback()` and pollutes
# the error message. This is cleaner:
f <- function() stop("Some letters are: ", head(letters), ".", call. = FALSE)
f()
#> Error: Some letters are: abcdef.

# For multiline messages consider the more readable `sprintf()` or `glue()`.
cat(sprintf("
  * Some letters: %s
  * Some months: %s.",
  toString(head(letters)), 
  toString(head(month.abb))
))
#> 
#>   * Some letters: a, b, c, d, e, f
#>   * Some months: Jan, Feb, Mar, Apr, May, Jun.

glue::glue("
  * Some letters: {toString(head(letters))}
  * Some months: {toString(head(month.abb))}.",
)
#> * Some letters: a, b, c, d, e, f
#> * Some months: Jan, Feb, Mar, Apr, May, Jun.

# Also consider `rlang::abort()` instead of `stop()`
# It doesn't have the annoying argument `call.`
# It makes it easy to create errors of specific class which facilitates testing
f <- function() rlang::abort("My error", class = "my_custom_error")
testthat::expect_error(f(), class = "my_custom_error")

# Sometimes it's nice to test the class of error regardless of the message
# because matchig the error string can be tricky
testthat::expect_error(f(), "My error.")
#> Error: `f()` threw an error with unexpected message.
#> Expected match: "My error."
#> Actual message: "My error"
```

<sup>Created on 2021-04-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

R/plot_timeline.R
```diff
@@ -0,0 +1,142 @@
...
+
+  if (!identical(sort(unique(lines_specs$line_name)), sort(unique(data$line_name)))) {
+    msg <- paste0(
+      "Can't find `line_name` values from 'lines_specs' in the data.\n",
+      paste0("* Unique `line_name` values in 'data' are: ",
+             toString(sort(unique(data$line_name))),".\n"),
+      paste0("* Unique `line_name` values in 'lines_specs' are: ",
+             toString(sort(unique(lines_specs$line_name))))
+    )
+    stop(msg)
```
maurolepore, line 70: Maybe extract all the checks so they don't bury the intent of the function. Something like this:

``` r
add_1 <- function(x) {
  check_add_1(x)
  
  # Do the main thing
  x + 1
}

check_add_1 <- function(x) {
  if (!is.numeric(x)) stop("`x` must be a number.", call. = FALSE)
  if (!identical(length(x), 1L)) stop("`x` must be of length 1.", call. = FALSE)
  # ... more checks.
}
```

<sup>Created on 2021-04-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

maurolepore (approved):
I noted a number of possible improvements.

README.Rmd
MonikaFu: I do want to keep the examples of how to use each plotting function in README. So a bit of repetition will be unavoidable I am afraid.

R/plot_timeline.R
MonikaFu: See commits 79db7a2 - 7a67e50.

R/plot_timeline.R
MonikaFu: See f1aa792

PR #62, opened 2021-03-29
7 mll snapshot

maurolepore:
Merges into 7-add-emissions-time-line to avoid modifying directly the original PR (idea by @AlexAxthelm).

This PR adds a snapshot of a .png file. It's the first time is use `expect_snaphsot_file()` but I'm pretty confident this is the way it's supposed to be used, as I basically copy-pasted the code in the helpfile.


cjyetman:
I have minimal experience using `expect_snaphsot_file()` in my still-private network plotting package. I use it to compare SVG outputs that I have to manually extract from D3 visualizations in htmlwidgets. I think for `{ggplot}` plots, `{vdiffr}` might be more ideal though because it can easily turn any object that has a print method into an SVG which it has nice visual comparison tools for. Either way, I'm interested in seeing how/what we ended using in this package and `{r2dii.interactive}`.

MonikaFu (approved):
LGTM. I like the idea of separate PRs. Makes me feel more in control of my own PRs :)

PR #79, opened 2021-04-14
(WAITING FOR #80) Support more R versions

### LINKED ISSUE #77: Support more R versions
Currently we support R >= 3.6. Potentially we could go further than that.

maurolepore:
Depends on #80 because removing dependencies might allow us to support older versions of R.

Closes #77
My approach here is to run `usethis::use_github_check_full()` to 

* update the gh workflow (to use the latest gh-action to run R CMD check on gh-actions).
* remove R versions that other r2dii packages on CRAN do not support (e.g. R 3.3).
* inspect the outcome of gh-actions to detect which R versions fail and why (e.g. because some dependency needs a higher version of R.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -29,12 +29,13 @@ jobs:
           - {os: macOS-latest,   r: 'release'}
           - {os: windows-latest, r: 'release'}
           - {os: windows-latest, r: '3.6'}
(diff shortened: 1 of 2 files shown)


cjyetman:
fyi....
https://github.com/2DegreesInvesting/r2dii.plot.static/pull/34#issuecomment-770754983
https://github.com/2DegreesInvesting/r2dii.plot.static/pull/34#issuecomment-770763235

maurolepore:
Thanks a lot CJ for pointing me to those comments. You just saved me a ton of time :-)

maurolepore:
Follow up at https://github.com/2DegreesInvesting/r2dii.plot.static/pull/81

PR #80, opened 2021-04-14
Prune dependencies

### LINKED ISSUE #78: Prune dependencies
Before releasing to CRAN we should get rid of as many dependencies as possible.

It is possible that we can already get rid of quite a few because of the moving of metareport functions into `r2dii.plot.internal` package.

maurolepore:
Closes #78 

This PR prunes needless dependencies, and those which are trivial to replace.

We may be able to remove a few more but with non-trivial changes -- so I would leave that for another PR. This focuses on quick wins.

--

Problematic dependencies (thanks CJ):

https://github.com/2DegreesInvesting/r2dii.plot.static/pull/34#issuecomment-770754983
https://github.com/2DegreesInvesting/r2dii.plot.static/pull/34#issuecomment-770763235


maurolepore:
```r
pkgnet::CreatePackageReport('r2dii.plot.static')
```

image

maurolepore:
Follow up in #85. Sorry for bothering.

PR #83, opened 2021-04-14
Dummy - Support more versions of R

maurolepore:


maurolepore:
Sorry for bothering everyone with so many PRs. It's hard to now do this when the very thing I need to debug is how the PR behaves on gh-actions.

PR #85, opened 2021-04-15
Prune dependencies and support move R versions

### LINKED ISSUE #78: Prune dependencies

maurolepore:
Closes #79 and closes #78 

This PR does two things because doing each separately caused me too much trouble. I don't know where the problem came from but I wasted several hours trying to chase a problem that never existed.

(If anyone wants to split this -- go ahead. I lost energy and run out of time today.)

1. Prune dependencies.
2. Update the gh workflow to the latest version (as of r-lib/actions) and to run R CMD check on as many versions of R as other r2dii packages.


MonikaFu (approved):
LGTM

PR #91, opened 2021-04-17
Address goodpractice::goodpractice()

maurolepore:
Relates to #66

Adress most `goodpractice::goodpractice()`﻿ issues, 
including adding tests.

`goodpractice` discovered three kinds of issues, and many issues of each kind. 

Warnings:

- Low test coverage.
- Use of `1:x` where `seq_along()` and friends would be safer.

Notes:

- Lines longer than 80 characters.

--

This PR is quite big but if you prefer I can try split this PR in a number of smaller PRs. If so, I would first merge the added tests, then the safer refactorings where I wrap lines at 80 characters, and finally the more sensitive `1:x` issues. Let me know.


R/plot_timeline.R
```diff
@@ -22,8 +22,10 @@
 #' )
 #'
 #' lines_specs <- dplyr::tibble(
-#'   line_name = c("projected", "corporate_economy", "target_demo", "adjusted_scenario_demo"),
-#'   label = c("Projected", "Corporate Economy", "Target Demo", "Adjusted Scenario Demo"),
+#'   label = c(
```
MonikaFu, line 25: This is not a good way to go.. The 'line_name' is really the important variable and the label is arbitrary. So if you want to derive one from the other, it should be the other way around..

tests/testthat/helper.R
```diff
@@ -11,3 +11,20 @@ fake_data <- function() {
...
+lines_specs <- function(...) {
+  dplyr::tibble(
+    label = c(
+      "Projected",
+      "Corporate Economy",
+      "Target Demo",
+      "Adjusted Scenario Demo"
+    ),
+    line_name = gsub(" ", "_", tolower(label)),
+    ...
```
MonikaFu, line 24: Same as in my previous comment. I find this implementation dangerous taking into account what the `lines_specs` represent.

tests/testthat/test-2dii_colours.R
```diff
@@ -0,0 +1,7 @@
+test_that("outputs the expected date", {
```
MonikaFu, line 1: Is this test message correct?

tests/testthat/test-plot_techmix.R
```diff
@@ -0,0 +1,71 @@
...
+
+  specs <- dplyr::tibble(
+    label = c(
+      "Portfolio 2020",
+      "Benchmark 2020",
+      "Portfolio 2025",
+      "Benchmark 2025",
+      "Target SDS 2025"
+    ),
+    metric_type = gsub(" ", "_", tolower(label))
```
MonikaFu, line 22: again, I find this implementation slightly illogical taking into account what `specs` are. At the same time this is only a test and it will work with the current dataset. However - it might be confusing. The `label` is a function of `metric_type` not the other way around.

tests/testthat/test-plot_techmix.R
```diff
@@ -0,0 +1,71 @@
...
+  )
```
MonikaFu, line 59: see my previous comment about specs

MonikaFu (changes requested):

R/plot_timeline.R
maurolepore: Okay, I reverted this change.

The reason why I did it is because I think the value is in the object, however it is created. And the way I created it here is the simplest one I could think of.

tests/testthat/test-plot_techmix.R
maurolepore: From `?plot_techmix` I learn that `df_bar_specs` is a "Dataframe containing order of bars and their labels (dataframe with columns: metric_type, label)". I don't see how it matters _hoe_ I create such data frame. Are we missing any piece of documentation that explains something I'm missing?

In any case, I prefer to move fast here so I'm quick to use the more straight forward approach of leaving the hard-coded values, as that duplication seems quite harmless.

tests/testthat/test-2dii_colours.R
maurolepore: Thanks! I fixed this title and all related titles so they are more relevant and consistent.

MonikaFu (approved):
LGTM. Thanks for addressing my comments!

PR #106, opened 2021-04-23
Enable ?r2dii.plot.static

maurolepore:
This PR adds a file for roxygen2 to automatically generate the package-wide helpfile -- which users can access with `?r2dii.plot.static`. This is a convenient way to learn about the package as a whole, and of accessing the index to see all features from inside R (hyperlink at the bottom of the helpfile).


MonikaFu (approved):
LGTM

PR #120, opened 2021-05-04
Add legend trajectory

### LINKED ISSUE #40: Create a custom legend for trajectory chart
Right now it is not possible to create an informative legend for the trajectory legend (explaining the areas and lines and how they relate to each other). Make that possible using R.

MonikaFu:
Closes #40 

In this PR I add a legend to the trajectory chart. 

In order to add a properly looking legend I need a couple of adjustments:
* I added a couple of helper functions. (Because I can't use a colourbar-type of legend with categorical variables for `fill` property I create a legend separately for a dummy chart which had the right look and add it to the final trajectory chart.) 
* I needed to adjust the colours in the main chart to reflect the palette that is created when using `scale_fill_stepsn()`.
* I simplified the way that the lines are added to the plot. 
* I removed the option to annotate the plot because it is not needed anymore. 

Resulting example plot:
image


R/plot_trajectory.R
```diff
@@ -176,70 +170,132 @@ plot_trajectory <- function(data,
...
+    ) +
+    theme(legend.position = NULL)
+
+  p_trajectory <- add_legend(p_trajectory,
+                             data_scenarios,
+                             scenario_specs,
+                            data_metrics,
+                              linetypes_ordered,
+                              linecolors_ordered,
+                              line_labels)
```
maurolepore, line 212: NB Style?

maurolepore (approved):
I see this is still in draft mode but as I'm here I'll add my review.

LGTM but notice:

* A test fails on multiple platforms and gives this hint 
> Run `snapshot_accept('plot_trajectory')` if this is a deliberate change

* I see odd indentation; you may run a styler.

MonikaFu:
@maurolepore I put it in the draft mode exactly because I wanted to spare you the time of looking at it before I talk to you about it. I know that the test is failing but the solution seems not to be as simple as updating snapshot (did that). Let's talk soon.

maurolepore:
Looks good to go. The errors I see on CI seem unrelated to our code.

PR #122, opened 2021-05-04
Simplify additional line metrics

### LINKED ISSUE #112: In plot_trajectory() handle additional_line_metrics more simply
The default to additional_line_metrics is data.frame(), and it's used
like so:

```r
  if (length(additional_line_metrics) >= 1) {
    ...
  }
 ```

This seems ovelry complicated. A simpler default might be TRUE, then
we could handle it like so:

```r
  if (additional_line_metrics) {
    ...
  }
```

MonikaFu:
Closes #112 

In this PR I simplify the way that `additional_line_metrics` is defined and used in code.

### DIFF
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -46,13 +46,11 @@
 #'   main_line_metric = main_line_metric
 #' )
 plot_trajectory <- function(data,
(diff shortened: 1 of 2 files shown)


maurolepore (approved):
LGTM

PR #124, opened 2021-05-05
Update snapshot of plot_trajectory

maurolepore:
For now we skip the test that deals with this snapshot, for all
platforms except ubuntu. I'm now on ubuntu so I take the chance to
update the snapshot.

I'm inclined to remove this snapshot completely, as it behaves very 
strangely (see also c3377ae).


maurolepore:
Given test-plot_trajectory.R is currenlty hard to maintain and
that this PR is only to my own benefit, I will 
- close this PR and check it out locally if I need it.  ﻿
- submit a separate PR deleting test-plot_trajectory.R

PR #125, opened 2021-05-05
Remove problematic snapshot test

maurolepore:
As is, this test file is hard to maintian and adds little
value (I'm the only one marginally using it). I think it
is not worth keeping it until the benefit justifies its
maintainability cost.﻿


MonikaFu (approved):
LGTM

PR #128, opened 2021-05-05
Extract labs() and refactor

maurolepore:
This PR simiplifies the implementation of `plot_timeline()`. It removes duplication and removes the code dealing with `labs()` to instead encourage users to use `ggplot2::labs()`.

The stored snapshot changed but I confirmed visually that the end product is as expected. Please confirm locally.

--

I suggest we discuss this PR and merge it -- then apply common changes to the other plotting functions.


R/plot_timeline.R
```diff
@@ -46,64 +46,53 @@
...
+  geom_line(
+    data = data,
+    aes(
+      x = .data$year,
+      y = .data$value,
+      colour = factor(.data$line_name, levels = lines_specs$line_name),
+      ...
+    )
+  )
 }
```
maurolepore, line 96: DRY common code between extrapolated/not-extrapolated lines.

MonikaFu:
LGTM. I run the code locally as well and the output looks as expected.

Issue #123, opened 2021-05-04
Remove plot and labels titles from plotting functions


MonikaFu:
As a maintainer of the code I would like the number of arguments of the plotting function to be as small as possible. 

I would like to arguments related to the plot title and x and y axis labels to be removed from the plotting functions and instead clear documentation to be provided on how to set the title yourself using the standard ggplot functions. 

In the end the user would call:
```
plot_r2dii() +
labs(
title = "Plot title",
x = "x axis title",
y = "y axis title"
)
```

PR #130, opened 2021-05-06
In plot trajectory remove labs

maurolepore:
Relates to #123 

This PR removes the code dealing with `labs()` and refactors to group code that relates to similar ggplot components, such as geoms, scales, themes, etc.


R/plot_trajectory.R
```diff
@@ -215,6 +209,12 @@ plot_trajectory <- function(data,
   p_trajectory
 }
 
+theme_trajectory <- function() {
+  theme(axis.line = element_blank()) +
+  theme(plot.margin = unit(c(0.5, 0.5, 0.5, 0.5), "cm")) +
+  theme(legend.position = NULL)
+}
+
```
MonikaFu, line 217: I'm not sure if this is necessary since it's just a couple of parameters and they are only used for this plot. I would leave them inside of the main function at the end of the plot specification.

MonikaFu (changes requested):

R/plot_trajectory.R
MonikaFu: I'll review it. When I used annotations I needed to change the margins for that. I don't anymore so it's possible that I could just remove this element.

MonikaFu: I will check it while refactoring trajectory chart.

MonikaFu (approved):
LGTM

PR #133, opened 2021-05-06
Refactor prepare_input_data()

maurolepore:
This PR refactors `prapare_input_data()` to capture my new understanding of what this function does:

1. add a new column `metric_type`
2. remove the prefix 'target_' from the existing column `metric`.


MonikaFu (approved):
LGTM

PR #138, opened 2021-05-07
In `plot_timeline()` simplify `specs` usage via new `timeline_specs()`

maurolepore:
This PR simplifies the usage of plot_timeline(). It exports
the new `timeline_specs()` which by default is created from
`data`. This gives us more control over how it's created and
simplifies the code we need to ensure its structure is
correct. 

The user can use the output of `timeline_specs(data)` as a
template to create their own specs, but I guess the default
should be fine most of the time.

I also simplified the tests. 
``` r
library(r2dii.plot.static)
library(tibble)

data <- tribble(
  ~year, ~line_name, ~value, ~extrapolated,
   2021,  "abc def",    0.1,         FALSE,
   2022,  "abc def",    0.2,         FALSE,
   2023,  "abc def",    0.3,         FALSE,
                                          
   2021,  "bcd efg",    0.5,          TRUE,
   2022,  "bcd efg",    0.6,          TRUE,
   2023,  "bcd efg",    0.7,          TRUE,
)

plot_timeline(data)
```

``` r
# `specs` default to 
timeline_specs(data)
#> # A tibble: 2 x 3
#>   line_name label   colour_hex
#>   <chr>     <chr>   <chr>     
#> 1 abc def   Abc Def #1b324f   
#> 2 bcd efg   Bcd Efg #00c082

# Users can edit this output to create custom `specs`. This is adapted from
# datapasta::tribble_paste(timeline_specs(data))
custom <- tibble::tribble(
  ~line_name,           ~label, ~colour_hex,
   "abc def", "Custom label A",   "#1b324f",
   "bcd efg", "Custom label B",   "#00c082"
  )

plot_timeline(data, specs = custom)
```

``` r
# Users may pick colours from our palette or elsewhere
r2dii_palette_colours()
#> # A tibble: 9 x 2
#>   label       colour_hex
#>   <chr>       <chr>     
#> 1 dark_blue   #1b324f   
#> 2 green       #00c082   
#> 3 orange      #ff9623   
#> 4 grey        #d0d7e1   
#> 5 dark_purple #574099   
#> 6 yellow      #f2e06e   
#> 7 soft_blue   #78c4d6   
#> 8 ruby_red    #a63d57   
#> 9 moss_green  #4a5e54
```

<sup>Created on 2021-05-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


MonikaFu:
@maurolepore - yes, that is exactly what we agreed on. Should we then create A/B versions of both functions? Or have A for one function and B for the other?

maurolepore:
See also https://github.com/2DegreesInvesting/r2dii.plot.static/pull/141#discussion_r628523891 where I suggest we might not deal with the legend at all and instead show users how to tweak it with `dplyr::recode()` or `dplyr::case_when()`.

maurolepore:
> Should we then create A/B versions of both functions? Or have A for one function and B for the other?

I prefer to be lazy so we can put the time we save elsewhere. I would have approach A in one function and approach B in another function; or A and B in only one function but not both -- maybe in `plot_timeline()` as it is the simplest one.

Also see my comment above. An alternative C would be to not deal with legend at all and show users how they can do it themselves with `dplyr::recode()`.

MonikaFu:
LGTM. Let's test with users which implementation of `specs` works best for them.

Issue #147, opened 2021-05-11
In prepare_for_*() let users specify which columns holds sectors?


maurolepore:
Follows from #145

>  It would be nice to be able to specify the column that the sectors are defined by, rather than being required to modify that column name in my data before passing to prepare_for_timeline()
-- @cjyetman

I move this to a separate issue so this wish doesn't get lost
when we close the more critical issues in #145.

PR #148, opened 2021-05-11
Enforce atomic sector and warn selection

### LINKED ISSUE #145: suggestions for prepare_for_timeline()
I was trying to use `prepare_for_timeline()` with investor data. Here are a few things I noticed/desired...

1. The `sector_filter` argument allows one to explicitly list the sectors they want, but I would like a way to specify *all* sectors without having to the each one out.
2. The `sector_filter` argument matches a set of sector names that do not exist in invest data (e.g. investor data has `Aviation` not `aviation`), so it's impossible to use without modifying all the sector names in my data to match these.
3. It would be nice to be able to specify the column that the sectors are defined by, rather than being required to modify that column name in my data before passing to `prepare_for_timeline()`

maurolepore:
Closes #145 (except for #147)

This PR enforces `sector_filter` of length 1 and enables using sector values in any letter case, e.g. "cement" and "Cement" are now valid both in the data and in the value passed via `sector_filter`.

I add a warning that tells the user which sector we select for them. This is intentionally annoying -- it is my way to suggest we should not filter at all, instead I suggest users use common functions such as `dplyr::filter()`, `subset()`.


maurolepore:
@cjyetman, let me know if this meets your general desire (except for #147). After your general approval I'll request Monika's review.

cjyetman:
I think the warnings are good/fine now, now that it's not the same message all the time no matter what. Maybe the warning for the first of a vector of sectors being chosen could give an indication of why one was chosen, e.g. "Can only use one sector. Using the first of the vector passed to `sector_filter`: automotive". Other than that, LGTM.... thanks!

MonikaFu:
I would be in favour here of adding a check on whether the user submitted only one 'sector' and if not throwing an error. For example as it is done in plot_techmix.R. Plus the description should be adjusted to inform the user correctly.

tests/testthat/test-prepare_for_timeline.R
```diff
@@ -105,7 +105,45 @@ test_that("with bad `extrapolate_missing_values` errors gracefully", {
...
+
+test_that("works with sectors in title case", {
+  data <- sda_target
+  data$sector <- tools::toTitleCase(data$sector)
+
+  expect_no_error(prepare_for_timeline(data, "Aviation"))
+
+  out <- prepare_for_timeline(data, "aviation")
+  expect_true(nrow(out) > 0L)
 })
```
maurolepore, line 149: @MonikaFu, if I understand correctly your suggestion is implemented via the error that `match.arg()` throws when it's passed a vector of length > 1. The test for it is in line 115 here. Also here are a few other tests for warnings. Assuming this is covered, and given CJ approved,  I'll go ahead and merge. If I'm missing something let me know and I'm happy to follow up in a separate PR.

cjyetman: My understanding of @MonikaFu's comment was that she preferred that if the user passed a vector of strings that was longer than one, that the function would stop with an error, not a warning, rather than arbitrarily choosing the first element sending a warning and then continuing. @MonikaFu please correct me if I'm wrong.

I don't have a really strong opinion about it, but that would be what I would lean toward. I can't really think of a strong justification for selecting the first element... and I think ultimately that trains users to do the wrong thing, rather than erroring out and telling them you have to select the sector first and filter your data to that. I can't imagine a scenario where a user would get what they wanted out of that unless it was by chance.

MonikaFu: Yes @cjyetman - that is exactly what I meant.

PR #154, opened 2021-05-14
Clarify and simplify tests

maurolepore:
This PR uses glue() and expect_snapshot_error() to
clarify and simplify error messages and tests. 

Along the way I realized that the tests or 
check_crucial_names() had a lot to improve so I did
the changes here. This is an internal function so I
would not worry too much about it. But I do want to 
test it produces the expected error message, then
use it checking not the message but the class of error.
This approach should make it easiest to maintain the
error message, so that if we change it we should get 
failing tests only that relate to check_crucial_names()
and not in all the functions that use it.


MonikaFu (approved):
LGTM

PR #158, opened 2021-05-15
In plot_timeline(), order legend to match lines

maurolepore:
On master, the order of the legend does not match the order of the
plot lines: 

Now they do

This is the recommended approach to plotting lines and legends:
https://twitter.com/JennyBryan/status/1244499728831672320?s=20

With this change, I we may not need to ask the user to provide
the order, making the interfaces simpler.

While I'm here, I refactored to simplify the body of `plot_timeline()`. The
snapshot changed but as far as I can tell the result is as expected.

Example:

``` r
devtools::load_all()
#> ℹ Loading r2dii.plot.static

specs <- tibble::tribble(
                ~line_name,           ~label, ~colour_hex,
               "projected",          "Proj.",   "#1b324f",
       "corporate_economy",  "Corp. Economy",   "#00c082",
             "target_demo",         "Target",   "#ff9623",
  "adjusted_scenario_demo",  "Adj. Scenario",   "#d0d7e1"
)



data <- prepare_for_timeline(
  extrapolate_missing_values = TRUE,
  sda_target,
  sector_filter = "cement",
  year_start = 2020,
  year_end = 2050,
  column_line_names = "emission_factor_metric",
  value_to_plot = "emission_factor_value"
)
plot_timeline(data, specs = specs)
```

``` r


data <- prepare_for_timeline(
  extrapolate_missing_values = FALSE,
  sda_target,
  sector_filter = "cement",
  year_start = 2020,
  year_end = 2050,
  column_line_names = "emission_factor_metric",
  value_to_plot = "emission_factor_value"
)
plot_timeline(data, specs = specs)
```

<sup>Created on 2021-05-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,15 +18,15 @@ Depends:
     R (>= 3.4)
 Imports: 
     dplyr,
+    forcats,
     ggplot2,
     ggpubr,
+    glue,
     lubridate,
     magrittr,
     r2dii.data,
     rlang,
-    scales,
-    ggpubr,
-    glue
+    scales
 Suggests: 
     covr,
     r2dii.analysis,
--- a/R/plot_timeline.R
+++ b/R/plot_timeline.R
@@ -48,21 +48,20 @@
 plot_timeline <- function(data, specs = timeline_specs(data)) {
   check_specs(specs, data)
 
+  data <- left_join(data, specs, by = "line_name")
+
   measured <- filter(data, !.data$extrapolated)
   plot <- ggplot() +
-    timeline_line(measured, specs) +
+    timeline_line(measured) +
     scale_x_date(expand = expansion(mult = c(0, 0.1))) +
     scale_y_continuous(expand = expansion(mult = c(0, 0.1))) +
     expand_limits(y = 0) +
-    scale_colour_manual(
-      values = specs$colour_hex,
-      labels = specs$label
-    )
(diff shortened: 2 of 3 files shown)


MonikaFu (approved):
LGTM! Good idea with the re-ordering of the legend. Part of my idea behind passing the order was to be able to give each label a colour that they choose. But it might be superfluous and unnecessary.

maurolepore:
Thanks Monika. This PR just improves the default. We can still work on the approach to pass custom order. I would suggest users use the ggplot2 approach so I added this to the examples of `plot_timeline()`:

``` r
library(ggplot2, warn.conflicts = FALSE)
library(dplyr, warn.conflicts = FALSE)
devtools::load_all()
#> ℹ Loading r2dii.plot.static

# Using default preparation and specs
data <- prepare_for_timeline(sda_target)
p <- plot_timeline(data)
p
```



``` r
# Customize as usual with ggplot2
custom <- c("red", "blue")
p +
  scale_colour_manual(values = custom) +
  labs(title = "Time line plot")
#> Scale for 'colour' is already present. Adding another scale for 'colour',
#> which will replace the existing scale.
```



<sup>Created on 2021-05-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

PR #163, opened 2021-05-19
In prepare_for_timeline() sector_filter is now obligatory 

maurolepore:
The argument `sector_filter` is now obligatory. Because we are now a bit more strict, some warnings are now gone.

This follows up #148. (Yet I think `prepare_for*()` functions should not do what the user can do with `dplyr::filter()`.)


MonikaFu (approved):
LGTM. Let us decide what the final solution for this will be when we get the feedback of the users.

PR #165, opened 2021-05-20
New versions of prepare_for_timeline*()

maurolepore:
This PR adds a simpler version of `prepare_for_timeline()`. For backward compatibility `prepare_for_timeline()` is identical to `prepare_for_timelineA()`, and the new flavour is called `prepare_for_timelineB()`.

`prepare_for_timelineB()`﻿ only extrapolates. So it's helpfile is simple and kind of boring. The complexity comes not from a complex single function but from the complex composition of multiple simple pieces, as shown in this example of the entire workflow:

``` r
library(tidyverse, warn.conflicts = FALSE)
devtools::load_all(quiet = TRUE)

data <- sda_target

data %>% 
  # Focus on a specific sector and time period with dplyr
  filter(sector == "cement", between(year, 2020, 2050)) %>%
  prepare_for_timelineB(extrapolate = TRUE) %>% 
  plot_timelineC(recode = toupper) +
  # Customize the plot with ggplot2
  scale_colour_manual(values = c("red", "blue", "green", "black")) +
  labs(title = "Timeline plot")
#> Scale for 'colour' is already present. Adding another scale for 'colour',
#> which will replace the existing scale.
```

<sup>Created on 2021-05-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


MonikaFu (approved):
LGTM. Let's test with the users and see which one we go with in the end. And as discussed - if we only keep the extrapolation then it should be renamed to `extrapolate()` or something like this and potentially added to the main plotting function.

PR #166, opened 2021-05-20
Add scenario colours

MonikaFu:
In this PR I add scenario colours data set to the package and use it in the trajectory plot so that the user does not need to specify the colours as part of specs.


MonikaFu:
@maurolepore - I've added the scenario colours data set to the package (it is called r2dii_scenario_colours) and I am trying to use it the `get_ordered_scenario_colours()` function in the plot_trajectory.R file but it seems like the dataset is not recognised in the function environment. Maybe you can help me here?

maurolepore:
* I defined the new dataset in `globalVariables()` to avoid R CMD check issues. 
* I fixed a typo in a column name (not `color` but `colour`). 
* I did other minor changes (redocument and ignore Rplots.pdf).

--

* It'd be nice to have at least some test for `plot_trajectory()`. 
* It'd be nice to figure out why Rplots.pdf keeps reappearing.

R/plot_trajectory.R
```diff
@@ -158,7 +157,44 @@ get_area_borders <- function(data) {
...
+  } else if (n == 4) {
+    nscenario_colours <- scenario_colours %>%
+           filter(.data$label %in%
+                    c("light_green", "dark_yellow", "light_yellow", "red"))
+  } else if (n == 5) {
+    nscenario_colours <- scenario_colours
+  } else {
+    rlang::abort(
+      glue(
+           "Scenario colours can be provided for between 1 and 4 scenarios. You provided {n - 1}."
```
maurolepore, line 178: Why between 1-4? The implementation works for `n` between 2-5. What am I missing?

R/plot_trajectory.R
```diff
@@ -158,7 +157,40 @@ get_area_borders <- function(data) {
...
+    "2" = pick(c("light_green", "red")),
+    "3" = pick(c("light_green", "light_yellow", "red")),
+    "4" = pick(c("light_green", "dark_yellow", "light_yellow", "red")),
+    "5" = assert_5_rows(r2dii_scenario_colours),
+    rlang::abort(glue(
+      "Scenario colours can be provided for between 1 and 4 scenarios. \\
+      You provided {n - 1}."
+    ))
+  )
+}
```
maurolepore, line 173: Here my refactoring using `switch()`.

R/plot_trajectory.R
```diff
@@ -158,7 +157,40 @@ get_area_borders <- function(data) {
...
+
+assert_5_rows <- function(data) {
+  stopifnot(nrow(data) == 5L)
+  invisible(data)
+}
```
maurolepore, line 178: `r2dii_scenario_colours` may change at some point. We need a way to know if it has more or less rows than 5.

R/plot_trajectory.R
```diff
@@ -158,7 +157,40 @@ get_area_borders <- function(data) {
...
+  switch(
+    as.character(n),
+    "2" = pick(c("light_green", "red")),
+    "3" = pick(c("light_green", "light_yellow", "red")),
+    "4" = pick(c("light_green", "dark_yellow", "light_yellow", "red")),
+    "5" = assert_5_rows(r2dii_scenario_colours),
+    rlang::abort(glue(
+      "Scenario colours can be provided for between 1 and 4 scenarios. \\
+      You provided {n - 1}."
+    ))
```
maurolepore, line 171: This function deals with `n`. Why `n - 1`?

R/plot_trajectory.R
MonikaFu: That is correct but n = 1 + number of scenarios (because of the 'worse' category).

R/plot_trajectory.R
MonikaFu: see my comment above

MonikaFu:
Thank you @maurolepore for refactoring! I consider this PR ready for review. I run the code locally and everything seems to be working correctly.

maurolepore (approved):
LGTM.

PR #167, opened 2021-05-21
update readme

maurolepore:
Follow up last PR.


maurolepore (commented):
Nothing exciting here -- just updating README as it seems to have fallen a bit behind the code.

PR #171, opened 2021-05-26
Update readme and move error 

MonikaFu:
This PR is a follow up on #166 which was merged too soon.

What it does:
* updates the README 
* Moves an error around the number of scenario colours to a more appropriate place


maurolepore (approved):
LGTM. I'm also happy for you to merge without review when the changes don't touch production code and fix a known issue.

MonikaFu:
@maurolepore - I changed some of your implementations and tests so I would appreciate you having a look at this PR now that it is ready for review. I also created a new issue #175 to add even more tests and checks for this function.

tests/testthat/test-plot_trajectory.R
```diff
@@ -1,11 +1,22 @@
...
-  expect_equal(nrow(get_ordered_scenario_colours(2)), 2)
-  expect_equal(nrow(get_ordered_scenario_colours(3)), 3)
-  expect_equal(nrow(get_ordered_scenario_colours(4)), 4)
-  expect_equal(nrow(get_ordered_scenario_colours(5)), 5)
-})
+test_that("with wrong number of scenarios errors gracefully", {
+  example_data <- process_input_data(example_data)
+
+  data <- prepare_for_trajectory_chart(
+    data = example_data, sector_filter = "power", technology_filter = "oilcap",
```
maurolepore, line 5: When I run `devtools::test()` I see:

```r
Warning (test-plot_trajectory.R:4:3): with wrong number of scenarios errors gracefully
partial argument match of 'data' to 'data_preprocessed'
```

You can use not `data = ...` but instead `data_preprocessed = ...`. However, I think it's best to rename the argument to just `data`, and the function to `prepare_for_trajectory()` -- the word "chart" adds little value.

--

Crazy idea for the record:

I would be even more succinct and use `prep_trajectory()` (but I remember you prefer to keep the `_for_`). We could even go a step more formal and define the "trajectory" class and add the constructor `new_trajectory()` ([as per this article](https://vctrs.r-lib.org/articles/s3-vector.html)). Then we could then implement `print.trajectory()` and call it as `print(new_trajectory(data))`.

maurolepore (approved):
LGTM. The only important change I suggest is to remove the warning I see when I call `devtools::test()`. See my comments below.

maurolepore:
I'd merge ASAP so that README works again and we
can run it from other PRs. 
﻿

MonikaFu:
I am going to merge this one is and we can address your comments about tests when we add more tests to this function.

PR #174, opened 2021-05-26
Add workflow to render all rmarkdown documents

### LINKED ISSUE #170: Add gh workflow to render readme

maurolepore:
Closes #170

﻿This should expose problems in .Rmd files, which may be out of sync with the production code. AFAIK, this workflow is fairly new and I'm not sure how robust it is but I think it's worth trying.

    usethis::use_github_action("render-rmarkdown.yaml")

    <https://github.com/r-lib/actions/tree/master/examples>

### DIFF
--- a/.github/workflows/render-rmarkdown.yaml
+++ b/.github/workflows/render-rmarkdown.yaml
@@ -1,0 +1,46 @@
+on:
+  push:
+    paths:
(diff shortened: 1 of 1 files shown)


MonikaFu (approved):
LGTM.

PR #176, opened 2021-05-26
Move process_input_data() into prepare_*() functions

### LINKED ISSUE #98: Should process_input_data(get_example_data()) become a dataset like sda_target?
process_input_data(get_example_data()) outputs a dataset, and sda_target
too. Should both live in data/? Or maybe both live in inst/extdata?

If there is a reason for them to be different kinds of data, then let's
discuss it. Else it would be nice to be consistent.

maurolepore:
Closes #98 

This PR simplifies the API by moving `process_input_data()` inside the two functions that call it. The function remains but is no longer exported.


MonikaFu (approved):
LGTM. Thia change was already discussed.

maurolepore:
I temporarily reverted this merge to reduce merge conflicts in other PRs. No action needed. I'll restore soon.

PR #183, opened 2021-05-28
Annotate lines instead of a legend

### LINKED ISSUE #177: Use annotations instead of legend in trajectory chart
Currently the code for the legend is very complex because we require a non-standard setup for the legend. 

As a package maintainer I would prefer the code to be simpler and more straightforward. Annotating the scenarios and lines directly in the plot might be a better solution.

MonikaFu:
Closes #177 

This PR replaces the complicated legend creation with annotations. In this way we are also able to retain the ggplot object properties and (for example) add custom labels to the plot. See example of a resulting plot below.

image


MonikaFu:
@maurolepore - I started this PR to have my work online but it is a draft and I am still going to do quite a bit of refactoring to simplify how the plot is created before I consider it ready for review.

maurolepore:
It looks like we need to `use_package("ggrepel")`.

https://github.com/2DegreesInvesting/r2dii.plot.static/pull/183/checks?check_run_id=2693143049

MonikaFu:
> It looks like we need to `use_package("ggrepel")`.
> 
> > Error: Error in loadNamespace(x) : there is no package called ‘ggrepel’
> 
> https://github.com/2DegreesInvesting/r2dii.plot.static/pull/183/checks?check_run_id=2693143049

Yes, definitely. I'm just fooling around for now. Once I decide on the packages I will add them to imports, etc.

R/plot_trajectory.R
```diff
@@ -95,33 +113,78 @@ plot_trajectory <- function(data,
...
+      ),
+      direction = "y",
+      color = "black",
+      size = 3.5,
+      alpha = 1,
+      nudge_x = 0.6,
+      nudge_y = 0.01 * value_span,
+      hjust = 0,
+      segment.size = 0.3,
+      xlim = c(min(data$year), last_year + 5)
```
cjyetman, line 165: Is this meant to `min(.data$year)` here, versus `min(data$year)`?

Maybe it's best to use a different argument name than `data` since it's so easily confused with `.data`? Maybe it doesn't make a difference? 🤷🏻 

here's an example of a case where it might matter...
``` r
library(dplyr)

data <- tibble::tribble(
  ~name,     ~gender, ~grade,
  "Bill",    "m",     65L,
  "Bob",     "m",     73L,
  "Mary",    "f",     98L,
  "Melissa", "f",     87L
)

data %>% 
  filter(gender == "f") %>% 
  mutate(mean(.data$grade))
#> # A tibble: 2 x 4
#>   name    gender grade `mean(.data$grade)`
#>   <chr>   <chr>  <int>               <dbl>
#> 1 Mary    f         98                92.5
#> 2 Melissa f         87                92.5

data %>% 
  filter(gender == "f") %>% 
  mutate(mean(data$grade))
#> # A tibble: 2 x 4
#>   name    gender grade `mean(data$grade)`
#>   <chr>   <chr>  <int>              <dbl>
#> 1 Mary    f         98               80.8
#> 2 Melissa f         87               80.8
```

MonikaFu: in this case I actually meant min(data) not min(.data)

maurolepore (approved):
LGTM -- great!

R/plot_trajectory.R
```diff
@@ -40,7 +40,8 @@
 #'
 #' p <- plot_trajectory(data,
 #'   scenario_specs_good_to_bad = scenario_specs,
-#'   main_line_metric = main_line_metric
+#'   main_line_metric = main_line_metric,
+#'   additional_line_metrics = additional_line_metrics
 #' )
```
maurolepore, line 45: I added `additional_line_metrics` because they were defined but not used.

PR #189, opened 2021-05-31
Use colours as data, and name them consistently

maurolepore:
This PR creates all "colours"﻿ datasets as data/ -- 
from data-raw/ and via usethis::use_data(). It also
names them consistently.

Do we need to export these datasets? Or can they be
for internal use only?

Does it make sense to move these datasets to r2dii.data?


maurolepore:
@MonikaFu, during our meeting today I forgot to ak if the `*_colours` datasets can be internal. If so, then we could keep them here and move them from data/ to R/sysdata.rda via `usethis::use_data(..., internal = TRUE)`.

PR #199, opened 2021-06-01
Colour datasets are now internal

maurolepore:
The datasets are now internal. This means they won't show up with `r2dii.plot::datasetname` but will still be available with `r2dii:::datasetname` (triple colon). Also, they are excluded from the website reference and index:

However, if we need to read the helpfile, we can still retrieve it with `?datasetname`, e.g. `palette_colours`:


MonikaFu (approved):
LGTM

PR #200, opened 2021-06-01
Add trajectory version B

### LINKED ISSUE #198: Add another version trajectory
Currently the user needs to specify the order of variables in specs document but it might be more convenient if they ordered the data frame itself using factors and that order would be used in the plotting function. 

Create a B-version of the function and ask users for feedback.

MonikaFu:
Closes #198 

In this PR I create another version of trajectory chart with simpler user interface (and requiring the user to order the data correctly before passing it to the function). See code below for use example:

```
# specify the order of metrics in data - first the main trajectory line, then
# the benchmarks and finally scenarios
lines_order <- c("projected", "corporate_economy", "sds", "sps", "cps")

data <- data %>%
  mutate(metric = factor(.data$metric, levels = lines_order)) %>%
  mutate(metric = dplyr::recode(
    .data$metric,
    "projected" = "Projeted",
    "corporate_economy" = "Corporate Economy",
    "sds" = "SDS",
    "sps" = "SPS",
    "cps" = "CPS")) %>%
  arrange(.data$year, .data$metric)

plot <- plot_trajectoryB(data)
plot
```


MonikaFu:
PR #201 is a cleaner version of this one.

PR #207, opened 2021-06-03
Refactor for consistency and polish documentation

maurolepore:
This PR makes the interfaces and helpfiles more consistent:

* All `prep_*()` functions now have arguments `value` and `metric`.
  I noticed that prep_techmix() and prep_trajectory() had `value` 
  (more precisely `value_to_plot`) but not `metric`. I think they
  should either both or none -- which we could achieve by expecting
  a specific column and asking users to rename their data if need to.

* Whenever possible, I document `@param`s in a single place and
  inherit the description from other helpfiles.

* The structure of `@param`s descriptions is now "<Type>. <Details>".

Also:

* I refactored including:
    - remove dead code (e.g. unused helpers).
    - remove live but needless code (e.g. some calls to `select()`).
    - Use standard check_<function_name>() to check `function_name()`.
    - Shorten names including arguments (e.g. `normalize` and `value`).

* I tested some untested code (e.g. that `sector_filter` has expected
  length).


tests/testthat/test-prep_techmix.R
```diff
@@ -47,11 +47,26 @@ test_that("with bad `years_filter` errors gracefully", {
...
       region_filter = "global",
       scenario_source_filter = "demo_2020",
       scenario_filter = "sds",
-      value_to_plot = "technology_share"
+      value = "technology_share"
+    )
+  )
+})
+
+test_that("with year_filter of length > 2 errors gracefully", {
```
MonikaFu, line 59: Actually in principle it could be possible to plot more than 2 years.

MonikaFu:
Overall LGTM although I would be keen on resolving the issues that I put in my comments.

PR #209, opened 2021-06-04
Move prep  functions into plot  functions

maurolepore:
This PR merges into #207 (not `master`). We could merge this one first, then the result into `master`.

This PR adds functions with minimal arguments. They are
named as `plot_*X()`. The goal is to show an extreeme
examples of how minimal the API can be.

I added `#' @examples` but didn't extend README yet. I'm happy to do so as we write articles to show the different interfaces.


R/plot_trajectoryX.R
```diff
@@ -0,0 +1,45 @@
...
+#' # `data` must meet documented "Requirements"
+#' data <- market_share %>%
+#'   filter(
+#'     sector == "power",
+#'     technology == "renewablescap",
+#'     region == "global",
+#'     scenario_source == "demo_2020"
+#'   )
+#' plot_trajectoryX(data, main = "projected")
+plot_trajectoryX <- function(data, main = "projected", normalize = TRUE) {
```
maurolepore, line 30: @MonikaFu, 
Notice that `plot_trajectoryX()` has the argument `main`. This implementation is naive: It only makes `main` the first element of the ordered set, and does nothing about the other elements of the set. Please see if this behaves as you expect or suggest/make changes.

MonikaFu (approved):
LGTM

PR #210, opened 2021-06-04
Add main_line argument trajectory

### LINKED ISSUE #205: plot_trajectoryB() might gain 'main_line'
Follows from https://github.com/2DegreesInvesting/r2dii.plot/pull/201#issuecomment-853135147

The excercise of developing `plot_trajectoryB()` exposed that the
most important information we need users to provide is the
'main_line', e.g. 'main_line = "projected"'. This is considerably
simpler than passing an ordered vector.

With this in mind, I wonder if prot_trajectoryB() should gain
the 'main_line' argument (how about calling it 'projected' instead?
or anything a bit more specific than 'main'?).

If I understand correctly, that would allow us to do the ordering inside
`plot_trajectoryB()` -- eliminating the risk that the user will forget
to order `metric`.

MonikaFu:
Closes #205 

In this PR I add a main_line argument to `plot_trajectoryB()`so that the user does not have to reorder data with `factor()` to indicate which data should be the main line. 

Example use:
``` r
devtools::load_all()
#> ℹ Loading r2dii.plot

data <- prep_trajectory(
  data = market_share, sector_filter = "power",
  technology_filter = "renewablescap",
  region_filter = "global", scenario_source_filter = "demo_2020",
  value = "production", end_year_filter = 2025,
  normalize = TRUE
)
plot_trajectoryB(data, main_line = "projected")
```

<sup>Created on 2021-06-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


maurolepore (approved):
LGTM! Awesome we can now get the order with so little input from the user. 

I have a few suggestions which I'll submit in a separate PR on top of this one. Along the way I'll do some refactoring.

PR #211, opened 2021-06-04
Extend #205 with my suggestions and refactorings

maurolepore:
Extend #205 with my suggestions and refactorings


MonikaFu (approved):
LGTM

PR #221, opened 2021-06-07
Fix annotations overlap

### LINKED ISSUE #203: Prevent scenario and trajectory annotations overlap
Currently if the trajectory lines are close to scenario lines there is a possibility for annotations overlap (see last comment of issue #172 ).

Possible solution: draw all annotations with one call to `greppel()` with conditional specification of the aesthetics.

MonikaFu:
Closes #203 

By drawing all annotations in one `ggrepel` call the issue of overlapping annotations is solved. See reprex for example.

Additional change introduced here: selection is added to `prep_trajectory()` as a quick fix for too many columns being returned otherwise when `normalise = TRUE`. 

``` r
devtools::load_all()
#> ℹ Loading r2dii.plot
library(tibble, quietly = TRUE)
library(dplyr, quietly = TRUE)
#> 
#> Attaching package: 'dplyr'
#> The following object is masked from 'package:testthat':
#> 
#>     matches
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

data <- prep_trajectory(
  market_share,
  sector_filter = "power", technology_filter = "oilcap",
  region_filter = "global", scenario_source_filter = "demo_2020",
  value = "production", end_year_filter = 2025,
  normalize = TRUE
)

scenario_specs <- tibble(
  scenario = c("sds", "sps", "cps"),
  label = c("SDS", "STEPS", "CPS")
)

main_line_metric <- tibble(metric = "projected", label = "Portfolio")
additional_line_metrics <- tibble(metric = "corporate_economy", label = "Corporate Economy")

# create overlapping data - replace portfolio with cps scenario
data_scen <- data %>%
  filter(.data$metric == "cps") %>%
  mutate(metric = "projected",
         metric_type = "portfolio")

data_overlap <- data %>%
  filter(.data$metric != "projected") %>%
  rbind(data_scen)

plot <- plot_trajectory(data_overlap,
  scenario_specs_good_to_bad = scenario_specs,
  main_line_metric = main_line_metric,
  additional_line_metrics = additional_line_metrics
)
plot
```

<sup>Created on 2021-06-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


maurolepore (approved):
LGTM except that the main line seems to be hidden behind a scenario line. I think the fix is to plot  first the scenario lines, then the benchmark, finally the portflolio -- as ggplot is a layered framework, the most visible line will be the one plotted last.

MonikaFu:
> LGTM except that the main line seems to be hidden behind a scenario line. I think the fix is to plot  first the scenario lines, then the benchmark, finally the portflolio -- as ggplot is a layered framework, the most visible line will be the one plotted last.

That is a good remark. I'll make an issue out of it as right now it's the opposite.

PR #239, opened 2021-06-08
Align X and Y articles

MonikaFu:
In this PR we plotting examples and customisations more so that it is clearly visible what are the differences between the two interfaces.


maurolepore (approved):
LGTM

MonikaFu:
> LGTM

Should we already merge this one in and adjust the other ones in a separate PR?

PR #241, opened 2021-06-09
Align X and Y articles for techmix and trajectory

MonikaFu:
In this PR I add further examples to articles in order to align the two versions.

There is also one change made in the `plot_trajectoryY()` code to align the two versions (the label is derived from data if not specified).

### DIFF
--- a/R/plot_trajectoryY.R
+++ b/R/plot_trajectoryY.R
@@ -57,7 +57,22 @@ plot_trajectoryY <- function(data,
                              scenario_specs_good_to_bad,
                              main_line_metric,
                              additional_line_metrics = NULL) {
+  abort_if_missing_names(scenario_specs_good_to_bad, "scenario")
   check_number_scenarios(scenario_specs_good_to_bad)
+  if (!("label" %in% names(scenario_specs_good_to_bad))) {
+    scenario_specs_good_to_bad$label = scenario_specs_good_to_bad$scenario
+  }
+
+  abort_if_missing_names(main_line_metric, "metric")
+  if (!("label" %in% names(main_line_metric))) {
+    main_line_metric$label = main_line_metric$metric
+  }
+  if (!is.null(additional_line_metrics)) {
+    abort_if_missing_names(additional_line_metrics, "metric")
+    if (!("label" %in% names(additional_line_metrics))) {
+      additional_line_metrics$label = additional_line_metrics$metric
+    }
+  }
 
   # plot scenario areas
   scenario_specs_areas <- get_ordered_scenario_specs_with_colours(
--- a/R/plot_trajectoryY.R
+++ b/R/plot_trajectoryY.R
@@ -60,17 +60,17 @@ plot_trajectoryY <- function(data,
   abort_if_missing_names(scenario_specs_good_to_bad, "scenario")
   check_number_scenarios(scenario_specs_good_to_bad)
   if (!("label" %in% names(scenario_specs_good_to_bad))) {
-    scenario_specs_good_to_bad$label = scenario_specs_good_to_bad$scenario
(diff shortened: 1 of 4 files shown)


R/plot_trajectoryY.R
```diff
@@ -57,7 +57,22 @@ plot_trajectoryY <- function(data,
                              scenario_specs_good_to_bad,
                              main_line_metric,
                              additional_line_metrics = NULL) {
+  abort_if_missing_names(scenario_specs_good_to_bad, "scenario")
   check_number_scenarios(scenario_specs_good_to_bad)
+  if (!("label" %in% names(scenario_specs_good_to_bad))) {
+    scenario_specs_good_to_bad$label = scenario_specs_good_to_bad$scenario
```
maurolepore, line 63: Style: We usually assign with `<-` not with `=`.

vignettes/articles/r2dii-plot-X-versus-Y.Rmd
```diff
@@ -91,8 +91,8 @@ caption <- "The X and Y APIs compared from a developer's perspective."
 knitr::kable(devs, caption = caption)
 ```
 
-To make the comparison concrete consider this small example. Notice the
-resulting plot is the same but the toolkit is different.
+To make the comparison concrete consider this small example of a trajectory plot (the other plot types you can find in the detailed ["X" API](https://2degreesinvesting.github.io/r2dii.plot/articles/articles/r2dii-plot-X.html)
+and ["Y" API](https://2degreesinvesting.github.io/r2dii.plot/articles/articles/r2dii-plot-Y.html)) articles. Notice the resulting plot is almost the same (except for the labels) but the toolkit is different.
```
maurolepore, line 95: NB: It's best to wrap at 80 characters so that the diffs on GH are clearer.

vignettes/articles/r2dii-plot-Y.Rmd
```diff
@@ -155,7 +216,7 @@ scenario_specs <- tibble(
 
 main_line_metric <- tibble(metric = "projected", label = "Portfolio")
 
-additional_line_metrics <- tibble(
+additional_line_metrics <- data.frame(
```
maurolepore, line 219: Wherever we use `data.frame()` I would use `stringsAsFactors = FALSE`. This is now the default but only since R 4.0 -- we still check on R < 4.0 and many users likely have R < 4.0 too. The old default `stringsAsFactors = FALSE` is a famous cause of trouble.

I would stick to `tibble()`. In general, I think that each package (e.g. r2dii.plot) is an opinion about how to do things. And that opinion may include a nudge about which external packages we recommend. DESCRIPTION already `Imports` and `Suggests` some packages that play well with `r2dii.plot` but we can also educate users via our documentation. If we need to show how a base-user would go about working with r2dii.plot, then I suggest we write a dedicated article doing everything with base, e.g. filtering with `subset()` or with `]`, mutating with `]<-`, reordering with `order()` and `factor()`, etc.

maurolepore (approved):
LGTM, thanks! -- my comments are NB. I may so minor changes myself then merge.

R/plot_trajectoryY.R
maurolepore: Solved via #242

vignettes/articles/r2dii-plot-X-versus-Y.Rmd
maurolepore: Solved via #242

vignettes/articles/r2dii-plot-Y.Rmd
MonikaFu: this change was driven by a comment of Jacob in thread where he indicated that our statement that "Y" API does not require knowledge of `dplyr`. So I wanted to show that it is indeed not required to use tibble.

PR #244, opened 2021-06-09
Use only data.frame()

maurolepore:
I prefer `tibble()` over `data.frame()`, but more prefer `data.frame()` over mixing both -- so this PR sticks to only `data.frame()`.


MonikaFu (approved):
LGTM

PR #245, opened 2021-06-09
DRY intro to r2dii.plot X and Y

maurolepore:
This PR reuses the content of the 'versus' article in README. The approach is to extract the common text into inst/intro.Rmd and call it from both places with `child = <path>`.

This way README is again useful and in sync with what we are doing elsewhere, and the duplication is cero. The articles X and Y serve as the integration test that README used to be.


maurolepore:
Not sure what's going on but I'm getting consistently stuck in `git remote hell`. Getting out seems more costly than resubmitting another PR so I'll do that and close this one. Sorry.

PR #252, opened 2021-06-09
Polish DESCRIPTION Title and Description

### LINKED ISSUE #249: Write package title and description like other r2dii packages
The DESCRIPTION file of other r2dii packages has been
intensively reviewed by a number of colleagues to
ensure the title and description reflect the product
clearly and accurately. I have shared with Monika a 
link to a google doc where this information is visible
for all r2dii packages. Once that doc is reviewed we
need to move the text to DESCRIPTION.

maurolepore:
Closes #249

﻿

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,16 +1,21 @@
 Package: r2dii.plot
-Title: Create Standard '2dii' Plots
+Title: Visualize Climate Scenario Alignment of Corporate Lending
+    Portfolios
 Version: 0.0.0.9000
 Authors@R: 
     person(given = "Monika",
            family = "Furdyna",
            role = c("aut", "cre"),
-           email = "monika.furdyna@gmail.com")
-Description: Allows the user to create standardized '2dii' graphs, like
-    trajectory charts or technology mix charts, and outputs them as
-    'ggplot' objects. It also contains data processing functions that
-    prepare the standard PACTA data to be an input for the plotting
-    functions.
+           email = "maurolepore@gmail.com")
+Description: These tools help you to assess if a corporate lending
+    portfolio aligns with climate goals. They visualize key climate
+    indicators attributed to the portfolio (e.g. production, emission
+    factors), and calculate alignment targets based on climate scenarios.
+    They implement in R the last step of the free software 'PACTA' (Paris
+    Agreement Capital Transition Assessment;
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -1,16 +1,21 @@
 Package: r2dii.plot
-Title: Create Standard '2dii' Plots
+Title: Visualize Climate Scenario Alignment of Corporate Lending
```
MonikaFu, line 2: I'd rather say 'Financial portfolios' as we're planning to move beyond banks' data as the next step

DESCRIPTION
```diff
@@ -1,16 +1,21 @@
...
            family = "Furdyna",
            role = c("aut", "cre"),
-           email = "monika.furdyna@gmail.com")
-Description: Allows the user to create standardized '2dii' graphs, like
-    trajectory charts or technology mix charts, and outputs them as
-    'ggplot' objects. It also contains data processing functions that
-    prepare the standard PACTA data to be an input for the plotting
-    functions.
+           email = "maurolepore@gmail.com")
+Description: These tools help you to assess if a corporate lending
```
MonikaFu, line 10: Again - financial instead of 'corporate lending' but maybe we should ask some analysts how to phrase it even better.

DESCRIPTION
```diff
@@ -1,16 +1,21 @@
...
+    portfolio aligns with climate goals. They visualize key climate
+    indicators attributed to the portfolio (e.g. production, emission
+    factors), and calculate alignment targets based on climate scenarios.
+    They implement in R the last step of the free software 'PACTA' (Paris
+    Agreement Capital Transition Assessment;
+    <https://2degrees-investing.org/>). Financial institutions use 'PACTA'
+    to study how their capital allocation decisions align with climate
```
MonikaFu, line 17: 'to visualise' instead 'to study'?

DESCRIPTION
maurolepore: Thanks. I'll accept your suggestions and copy them to the google doc for broader feedback.

DESCRIPTION
maurolepore: Mmm ... here we are talking not about why financial institutions use this package but about why they use  'PACTA' -- so we need a verb that is broader than the scope of this package. I think "study" is a pretty good fit and is exactly as the description in other r2dii packages -- see the related google doc. I lean towards leaving it like it is.

PR #256, opened 2021-06-14
Make the X API the main API

maurolepore:
This PR follows from feedback after beta testing
the X and Y APIs. Here I keep the X API and
remove the Y API. I do not change behaviour 
except for adding hints to existing error messages.

Now all files are covered by tests by 75% or more.
The uncovered code seems easy to cover and it
may be worthwhile covering since there is still room
for a lot more refactoring. But that may well follow in 
another PR, as this one is already huge.


R/plot_techmixX.R
```diff
@@ -24,11 +24,224 @@
...
+    c(
+      "metric",
+      "region",
+      "scenario_source",
+      "sector",
+      "technology",
+      "technology_share",
+      "year"
+    )
+  }
```
maurolepore, line 41: Ups! I forgot to remove this and instead reuse a duplicated helper.

R/plot_techmixX.R
```diff
@@ -24,11 +24,224 @@
...
+
+  hint_if_missing_names(
+    abort_if_missing_names(data, crucial_techmix_and_trajectory())
+  )
+
+  abort_if_has_cero_rows(data)
+  abort_if_multiple(data, "sector")
+  abort_if_multiple(data, "region")
+  abort_if_multiple(data, "scenario_source")
```
maurolepore, line 50: This is one example of stuff I need to refactor. I may do it here or follow up in a separate PR.

R/plot_techmixX.R
```diff
@@ -24,11 +24,224 @@
...
+
   prep <- prep_techmixB(data)
+
   found_scenarios <- pull_scenarios(prep)
   abort_if_invalid_length(found_scenarios)
 
   plot_techmixY(prep)
 }
+
+prep_techmixB <- function(data, value = "technology_share", metric = "metric") {
```
maurolepore, line 60: The suffix "B" and similar no longer makes sense so I may rename to simply `prep_*()`.

R/plot_techmixX.R
maurolepore, line 50: This may become a call to `lapply()` but first I need to add an `env` argument.

R/utils.R
```diff
@@ -90,3 +91,86 @@ abort_if_multiple <- function(data, x) {
...
+  invisible(expr)
+}
+
+format_plot_function_name <- function(.expr) {
+  # Matches "fun_name" from "fun_name(...)"
+  fun <- gsub("(.*)\\(.*", "\\1", .expr)
+  # Matches "_name" from "fun_nameZ"
+  fun <- gsub(".*_(.*)[A-Z]", "\\1", fun)
+  fun <- glue("plot_{fun}")
+}
```
maurolepore, line 128: I think this is dead code.

tests/testthat/test-plot_trajectoryX.R
```diff
@@ -1,3 +1,33 @@
...
+
+  data <- market_share %>%
+    filter(
+      sector == first(sector),
+      technology == first(technology)
+    )
+  p <- plot_trajectory(data)
+  p$plot_env <- NULL
+  expect_snapshot(str(p))
+})
```
maurolepore, line 13: Snapshot testing gg objects is brittle. I use this only on my computer to ensure I'm not changing behaviour while I refactor. But I may remove this once I'm done with the most aggresive refactoring.

tests/testthat/test-plot_trajectoryX.R
```diff
@@ -7,6 +37,171 @@ test_that("outputs a ggplot", {
...
+    )
+  out <- prep_trajectoryB(data)
+
+  skip("Dead code")
+  expect_snapshot(out)
+})
+
+test_that("with multiple distinct values in some columns errors gracefully", {
+  long_sector <- mutate(head(market_share, 2), sector = 1:2)
+  skip("Dead code")
```
maurolepore, line 123: I need to prune tests and also to make them work with less data for speed (e.g. `data <- head(sda)` instead of `data <- sda`).

maurolepore (commented):
I made some note to myself about things I may add to this PR.

maurolepore:
@MonikaFu, this is still a draft mostly because I noted a few things I could add. But it's quite okay as it is. Note this PR touches most files so if you are working on something new you may PR off this one or wait until we merge into `master`.

R/plot_techmixX.R
maurolepore: Done for the `prep_` functions -- not for the `plot_` funcitons. Because of potential clashes, that will come after more test coverage and aggressive refactoring -- so another PR.

R/utils.R
maurolepore: Nope, it's alive.

maurolepore:
@MonikaFu this is ready for review. This PR should not change the behaviour of the X API except for improved error messages. Also the names of the main functions no longer use the suffix X (or similar). 

There are two places I would like you to inspect to see if we have dead code or something we need to test.

In separate PRs (and ASAP) we need to incorporate feedback (e.g. rename to `plot_emissions_intensity()`) and add the improvements you want to make. I will mostly focused on refactoring unless you need something else.

R/plot_timeline.R
```diff
@@ -0,0 +1,140 @@
...
+
+# TODO: Is this dead code? How can we test it?
+abort_if_too_many_lines <- function(data) {
+  n_lines <- nrow(data)
+  n_colours <- nrow(palette_colours)
+  if (n_lines > n_colours) {
+    abort(glue(
+      "Can't plot more than {n_colours} lines. Found {n_lines} lines:
+        {toString(data$line_name)}."
+    ))
```
maurolepore, line 127: Monika?

R/plot_trajectory.R
```diff
@@ -428,6 +227,7 @@ get_area_borders <- function(data) {
         start_value_portfolio - perc_distance_upper_border * value_span
       value_span <- upper_area_border - lower_area_border
     } else {
+      # TODO: Is this dead code? How can we test it?
       upper_area_border <-
         perc_distance_lower_border * value_span + start_value_portfolio
       value_span <- upper_area_border - lower_area_border
```
maurolepore, line 233: Monika?

MonikaFu (commented):
I checked out the PR and run locally and both trajectory and techmix charts look good (I am going to implement some improvements once this PR is merged). However, for the timeline I believe that we should use the Y version in which the interface is equally simple but the line labels are guessed and modified which gives them better look without involving the user (see the first code snippet in the [Y article](https://2degreesinvesting.github.io/r2dii.plot/articles/articles/r2dii-plot-Y.html).

Once we change this I think that the PR will be ready for merging.

R/plot_techmix.R
```diff
@@ -0,0 +1,163 @@
...
+#'   )
+#' plot_techmix(data)
+plot_techmix <- function(data) {
+  stopifnot(is.data.frame(data))
+  hint_if_missing_names(
+    abort_if_missing_names(
+      data, c(common_crucial_market_share_columns(), "technology_share")
+    )
+  )
+  abort_if_has_cero_rows(data)
```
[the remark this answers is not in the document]
maurolepore: Sorry I don't see the typo-candidate. Which word looks misspelled?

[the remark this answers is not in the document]
maurolepore: Oh [z]ero! Yes, that's a typo I didn't see because in Spanish it's [c]ero. Thanks!

R/plot_trajectory.R
maurolepore: Great. No problem or rush then. I just fail to figure out a test that would touch that line. If you do think of such test please share it. Same with the other similar comment.

R/plot_timeline.R
maurolepore: Thanks! What would be an example of "too many lines"? Do you mean the case when `data$metric` has more than 6 unique values? Or the case when the number of scenarios (e.g. "target_sds", "target_cps", ...). is greater than 4?

I'm trying to come up with a test that would touch this line but I fail to figure it out. Sometimes code dies because an argument is no longer used so a chunk is never triggered -- and can be safely removed if the chunk is re-implemented upstream.

maurolepore:
>  For the timeline I believe that we should use the Y version in which the interface is equally simple but the line labels are guessed and modified which gives them better look without involving the user (see the first code snippet in the Y article.

Great, let's see if I understand. `plot_timelineX()` (now `plot_timeline()`) internally calls `plot_timelineB()` (now `plot_timeline_impl()`) but instead it should call `plot_timelineY()`. Right?

MonikaFu (approved):
@maurolepore - should we fix the typo? [c]ero ? Otherwise LGTM

PR #287, opened 2021-06-17
plot_techmix() now outputs legend labels like the other plots

maurolepore:
This PR might be a bit controversial. It ugly-fies the legend
labels of plot_techmix() so they look like the ones of the
other two plots. This is for consistency and so we can later
give users the ability to customize the legends in the same
way.


AlexAxthelm:
@MonikaFu , is this okay with you? it looks like it's the only open PR at the moment, and then we can send it to CRAN?

MonikaFu:
@AlexAxthelm I haven't put my thoughts online yet but it actually is not fully understandable for me why we'd do this so I want to discuss with @maurolepore before approving. As for CRAN - as far as I understood he wanted to make some more improvements/pruning before we submit to ensure positive outcome. We should check with him once he is online. I also have one more change to do before we submit.

PR #290, opened 2021-06-17
Filter data to startyear

### LINKED ISSUE #275: FIXME: plot_trajectory() with year 2019 errors ungracefully
Via the entire r2dii workflow I arrived to a dataset that looks like the one below. Notice one row where year is 2019.
With this dataset `plot_trajectory()` throws an error which message is uninformative. We need to find out what's the
problem and help the user fix it (of fix it ourselves).

``` r
devtools::load_all()
#> ℹ Loading r2dii.plot

  data <- tibble::tribble(
  ~sector,     ~technology, ~year,  ~region, ~scenario_source,             ~metric,     ~production, ~technology_share,
  "power", "renewablescap", 2019L, "global",      "demo_2020", "corporate_economy",               1,               0.1,
  "power", "renewablescap", 2020L, "global",      "demo_2020", "corporate_economy",               1,               0.1,
  "power", "renewablescap", 2020L, "global",      "demo_2020",         "projected",               1,               0.1,
  "power", "renewablescap", 2020L, "global",      "demo_2020",        "target_sds",               1,               0.1,
)

plot_trajectory(data)
#> Error in if (delta_distance > max_delta_distance) {: argument is of length zero
```

<sup>Created on 2021-06-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

MonikaFu:
Closes #275 

In this PR I add the functionality to filter the plotting data year to start of 'projected' to graphs: trajectory, techmix.


tests/testthat/test-plot_techmix.R
```diff
@@ -64,3 +64,27 @@ test_that("with missing crucial names errors gracefully", {
   bad <- select(data, -scenario_source)
   expect_error(class = "hint_missing_names", plot_techmix(bad))
 })
+
+test_that("with input data before start year of 'projected' prep_techmix outputs data with start year of 'projected'", {
```
maurolepore, line 68: NB: Maybe wrap at 80 characters

tests/testthat/test-plot_techmix.R
```diff
@@ -64,3 +64,27 @@ test_that("with missing crucial names errors gracefully", {
...
+  data <- subset(
```
maurolepore, line 69: Just be aware that `subset()` is not safe to program with.

`?subset`
> WARNING
> This is a convenience function intended for use interactively. For programming it is better to use the standard subsetting functions like [, and in particular the non-standard evaluation of argument subset can have unanticipated consequences.

`dplyr::filter()` is also a non-standard evaluation function but it's made safe to program with via the tidyeval framework. In general, if there is a tidyverse equivalent for a base R function that is for a reason -- with the tidyverse option being the safest one (e.g. `data.frame(stringsAsFactors = TRUE)` (in R < 4) vs `tibble::tibble()`). That's why I generally prefer to show examples with tidyverse functions.

maurolepore (approved):
LGMT

tests/testthat/test-plot_techmix.R
MonikaFu: I see. I used it because I saw you using it in all other tests. So what is your advice here?

maurolepore: Thanks for pointing that out. #302 removes all use of `subset()` except for examples as it's what @AlexAxthelm prefers.

PR #328, opened 2021-06-21
Revert colours automotive

### LINKED ISSUE #140: Revert colours automotive - techmix
Currently for power the order of colour intensity is from dark (brown tech) to light (green tech) but the opposite for automotive. This should be aligned.

MonikaFu:
Closes #140 

In this PR the automotive colours are reverted so that the darkest corresponds to the least sustainable technology (as in power). I also reversed the order of colours in techmix legend so that it follows that in the bars.


MonikaFu:
This is a small change that only affects that the visual output. Merging without a review.

PR #329, opened 2021-06-21
Report latest user-facing NEWS

maurolepore:
This PR follows up #328. From now on we need to keep track
of user-facing changes in NEWS.md, and polish it just
before the next release. Changes that do not face users
(e.g. refactoring) do not need to be reported﻿ (although
sometimes I patch a version with only internal changes and
report "* Maintenance release" in NEWS.


MonikaFu (approved):
LGTM

PR #339, opened 2021-06-22
Refactor plot emission intensity

### LINKED ISSUE #263: Refactor to remove prep_*() functions
As a maintainer I want to see right away what data manipulations are performed on the data before passing it to plot_*().

Context: currently most of the data processing has been moved onto the user via `dplyr::filter()`. The `prep_*()` functions became very thin and seem to unnecessarily obscure the data manipulations needed for the plots. Since we decided to fix the plots more closely to work only with banks data for now some processing steps are not necessary and the code could work with the output of r2dii.analysis directly. 

Solution: move part of the `prep_*()` functions steps into separate helper functions (like normalize) and move the rest into the plotting functions itself.

maurolepore:
Closes #263 
Closes # 324

- Refactor.
- The only user-facing change should be improved error messages.


R/plot_trajectory.R
```diff
@@ -22,164 +22,140 @@
...
+    direction = "y",
+    color = "black",
+    size = 3.5,
+    alpha = 1,
+    nudge_x = if_else(is_scenario(lines_end$metric0), 0.6, 0.1),
+    nudge_y = 0.01 * value_span(data),
+    hjust = 0,
+    segment.size = if_else(is_scenario(lines_end$metric0), 0.4, 0),
+    # ASK: Does `6` have a meaning? e.g. `some_space <- 6`. I changed the
+    # value to 1-30 and noticed no effect on the plot. Strange.
```
MonikaFu, line 82: It's a limit so it will not necessarily impact the plot. The 6 is arbitrary and I actually need to change it something that is varying depending on the length of the time interval we are plotting. The point was basically to have a ratio of annotation space to plot space 1:4

R/plot_trajectory.R
```diff
@@ -22,164 +22,140 @@
...
-      nudge_y = 0.01 * value_span,
-      hjust = 0,
-      segment.size = if_else(data_lines_end$metric_type == "scenario", 0.4, 0),
-      xlim = c(min(data$year), last_year + 6)
-    ) +
-    scale_fill_manual(
-      aesthetics = "segment.color",
-      values = line_colours
-    )
+    # ASK: We call `scale_fill_manual()` twice with `value`. I don't get it.
```
MonikaFu, line 88: they correspond to different objects and it is possible because I use the `aesthetics = "segment.color"` for the second one. First one is for the coloured areas and the second one is about the lines going from the plot to the annotation label (the "segments").

MonikaFu (approved):
LGTM

PR #341, opened 2021-06-22
MERGE FIRST. Refactor plot_trajectory()

maurolepore:
This PR merges into `Refactor_plot_techmix()`. It is mostly refactoring, 
and is a piece of a much larger PR that I decided to break down.﻿


R/plot_trajectory.R
```diff
@@ -246,58 +226,51 @@ get_ordered_scenario_colours <- function(n) {
...
   )
 }
 
-prep_trajectory <- function(data,
-                            value = "production",
-                            metric = "metric") {
-  cols <- c("year", "metric_type", "metric", "technology", "value")
+prep_trajectory <- function(data) {
+  # Store original values to detect metric type from metric
+  data$metric0 <- data$metric
```
MonikaFu, line 235: I'm not fully convinced with this naming. To me it is harder to understand `metric` and `metric0` compared to `metric` and `metric_type`. But I am thinking of abandoning `metric_type` and working directly on `metric` with helpers. And then have a column for labels which is called `label` (to me it would be easier to follow as a user). So we can leave it like this for now. Let's discuss my idea later.

tests/testthat/_snaps/plot_trajectory.md
```diff
@@ -5,50 +5,46 @@
 # if `data` is not market_share-like errors gracefully
 
     `data` must have all the expected names.
-    x Missing names: metric, region, scenario_source, technology.
+    x Missing names: metric, production, region, scenario_source, technology.
```
MonikaFu, line 8: is this correct? I guess that 'production' is one of the metrics? So then I think this message might be confusing to the user.

MonikaFu (approved):
LGTM but please see my comments.

R/plot_trajectory.R
maurolepore: I agree. I prefer 

* `metric -> label` or ` -> legend_label` (to disambiguate from, say, axis label).

* `metric0 -> metric`. It's useful to leave inputs untouched. For example, `data$metric` has "target_sds" and we change it to "SDS", then error messages using that column would mention "SDS" which may leave the user wondering where does "SDS" come from. The principle of lease surprise guides us to throw messages with "target_sds".

--

 Notice that  `plot_*()` returns no data so the names of the columns are only exposed to us programmers -- not users. What we do expose are the values of those columns, via error messages as legend labels.

maurolepore: See #342

tests/testthat/_snaps/plot_trajectory.md
maurolepore: Not sure if I understand your comment. `production` is a column of `market_share`

``` r
hasName(r2dii.plot::market_share, "production")
#> [1] TRUE
```

<sup>Created on 2021-06-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

PR #346, opened 2021-06-22
Clarify the use of aesthetics in scale_fill_manual

maurolepore:
Here I'm using the default value explicitely to make it clearer
why we call scale_fill_manual() twice.

This is not strictly necesary but should stop me from trying to 
remove one call to scale_fill_manual().


MonikaFu (approved):
LGTM

PR #364, opened 2021-06-25
Add qplots for techmix and emission intensity

MonikaFu:
In this PR I add quick plots for:
* techmix
* emission intensity

Also - some refactoring is performed.


R/plot_techmix.R
```diff
@@ -70,34 +70,53 @@ abort_if_multiple_scenarios <- function(data, env = parent.frame()) {
...
+  }
+
+  start_year <- min(out$year)
+  future_year <- max(out$year)
+  if (!quiet()) {
+    inform(glue(
+      "The `technology_share` values are plotted for extreme years.
+       Do you want to plot different years? E.g. filter the data with:\\
+       `subset(data, year %in% c(2020, 2030))`."
+    ))
```
MonikaFu, line 97: I am not sure if this is not an overkill.

R/plot_techmix.R
```diff
@@ -70,34 +70,53 @@ abort_if_multiple_scenarios <- function(data, env = parent.frame()) {
...
 
 plot_techmix_impl <- function(data) {
-  colours <- semi_join(technology_colours, data, by = c("sector", "technology"))
-  metrics <- rev(unique(data$metric))
+  colours <- semi_join(technology_colours, data, by = c("sector", "technology")) %>%
+    left_join(
+      data %>%
+        select(.data$technology, .data$label_tech) %>%
+        unique(),
+      by = "technology")
```
MonikaFu: I need the `label_tech` from 'data'. If there is any cleaner way to get it - I'd like to know!

R/plot_emission_intensity.R
```diff
@@ -32,28 +32,37 @@ check_plot_emission_intensity <- function(data, env = parent.frame()) {
...
-    ) +
+  ggplot(
+    data = data,
+    aes(
+      x = .data$year,
+      y = .data$emission_factor_value,
+      colour = match_lines_order(data)
+    )
+  ) +
+    geom_line() +
```
maurolepore, line 65: I like that `ggplot()` holds useful `data` so we can inspect it in tests with `p$data` rather than sometimes via `p$data`, and other times via `p$layers[[1]]$data`, `p$layers[[2]]$data`,  or `p$layers[[n]]$data`.

R/plot_techmix.R
```diff
@@ -70,34 +70,56 @@ abort_if_multiple_scenarios <- function(data, env = parent.frame()) {
...
       value = .data$technology_share,
-      sector = recode_sector(.data$sector)
+      sector = recode_sector(.data$sector),
+      label = convert_label(.data$label),
+      label_tech = convert_tech_label(.data$technology)
     )
+
+  if (span_5yr) {
+    out <- span_5yr(out)
+  }
```
maurolepore, line 88: Same as before, `prep_common()` could the the single place where we compute `convert_label` and `span_5yr`.

R/plot_techmix.R
maurolepore: Do you intend to inform both when `span_5yr = TRUE` and `span_5yr = FALSE`?

maurolepore: Should we enhance the message with the user's data? E.g. 

* replacing the literal "extreeme years" with the actual years we are plotting.
* replacing `data` with the symbol the user used to name their data -- see how by searching for calls similar to `.data <- deparse_1(substitute(data, env = env))`
* replacing `"... c(2020, 2030) ..."` with `"... {range(data$year, na.rm = TRUE)} ..."`

R/plot_techmix.R
```diff
@@ -70,34 +70,56 @@ abort_if_multiple_scenarios <- function(data, env = parent.frame()) {
...
-  colours <- semi_join(technology_colours, data, by = c("sector", "technology"))
-  metrics <- rev(unique(data$metric))
+  colours <- semi_join(technology_colours, data, by = c("sector", "technology")) %>%
+    left_join(
+      data %>%
+        select(.data$technology, .data$label_tech) %>%
+        unique(),
+      by = "technology"
+    )
+  labels <- rev(unique(data$label))
```
maurolepore, line 112: Maybe extract this into `techmix_labels()` so it doesn't obscure the structure of the plot? It then becomes

```r
labels <- techmix_labels(data)
ggplot(
  ...
```

R/qplot_techmix.R
```diff
@@ -0,0 +1,54 @@
...
+    labs(
+      title = glue("Current and future technology mix for the {sector} sector")
+    )
+}
+
+#' @examples
+#' format_label_techmix(c("corporate_economy", "target_sds"))
+#' # Weird case
+#' format_label_techmix(c("corporate_._economy", "target_sds_abc"))
+#' @noRd
```
maurolepore, line 49: These examples don't show off the `recode_metric` feature.

R/utils.R
```diff
@@ -254,6 +254,7 @@ span_5yr <- function(data) {
 
 to_pretty_label <- function(technology) {
   label <- to_title(technology)
+  label <- sub("^(?i)ice", "ICE", label)
   label <- sub("cap$", " Capacity", label)
   label <- sub("_hdv$", "Heavy Duty Vehicles", label)
   label
```
maurolepore, line 260: Maybe a more precise name would be `decode_tecnology()` or `spell_out_technology()`? Why do we not spell out `ICE` but we do spell out `cap` and `hdv`?

tests/testthat/_snaps/qplot_emission_intensity.md
```diff
@@ -0,0 +1,12 @@
...
+
+    Emission intensity trend for Automotive
+
+# Prints axis labels as expected
+
+    [1] "Year"
+
+---
+
+    [1] "Tons of CO2 per ton"
```
maurolepore, line 11: Tons of Co2 per ton <of what?>

tests/testthat/_snaps/qplot_techmix.md
```diff
@@ -0,0 +1,4 @@
+# Has the title as expected
+
+    Current and future technology mix for the Power sector
```
maurolepore, line 3: The letter case seems odd: It's sentence case until we reach "Power". A more conventional way to emphasize a word is with _italics_ or "quotes", or simply by making it take the emphasis position of a sentence -- the position right next to a colon, semi-colon, or period.

Suggestions:

* Current and future technology mix for the _power_ sector
* Current and future technology mix for the "power" sector 
* Current and future technology mix for the sector power.

tests/testthat/test-plot_emission_intensity.R
```diff
@@ -28,13 +28,13 @@ test_that("outputs an object with no factor-columns derived from `specs`", {
...
+test_that("doesn't output pretty labels", {
   data <- filter(sda, sector == "automotive")
   p <- plot_emission_intensity(data)
 
-  metrics <- unique(p$layers[[1]]$data$emission_factor_metric)
-  pretty <- c("Projected", "Corporate Economy")
-  expect_equal(pretty, metrics)
+  metrics <- unique(p$data$label)
+  ugly <- c("projected", "corporate_economy")
+  expect_equal(ugly, metrics)
```
maurolepore, line 37: I think we need to pass `ugly` to the `expected` argument of `expect_equal()`, i.e. 

```r
expect_equal(actual = metrics, expected = ugly)
```

This will result in clearer failure messages. See how actual and expected are mixed up here

image

maurolepore: Yes, I did this wrong in many places :-)

tests/testthat/test-qplot_emission_intensity.R
```diff
@@ -0,0 +1,33 @@
...
+})
+
+test_that("Prints axis labels as expected", {
+  data <- filter(sda, sector == "automotive")
+  p <- qplot_emission_intensity(data)
+
+  expect_equal(p$labels$x, "Year")
+  expect_snapshot_output(p$labels$x)
+
+  expect_match(p$labels$y, "[Tt]ons of CO2 per ton")
```
maurolepore, line 24: I often use `[Aa]` because I want to ignore the letter case of A. But in this case we likely want this test to fail if "A" becomes "a", right?

tests/testthat/test-qplot_techmix.R
```diff
@@ -0,0 +1,68 @@
...
+  data <- filter(
+    market_share,
+    sector == "power",
+    region == "global",
+    year <= 2025,
+    metric %in% c("projected", "corporate_economy", "target_sds")
+  )
+  p <- qplot_techmix(data)
+
+  expect_true(max(p$data$year) - min(p$data$year) <= 5)
```
maurolepore, line 11: Do we need `na.rm = TRUE`? We haven't tested the package when `data` has `NA`, so I fear we'll see bugs where we call `min()`, `max()`, `range()`, `sum()`, and every other function that includes `na.rm`.

maurolepore (approved):
LGTM. I think all my comments are NB.

R/plot_techmix.R
maurolepore: For messages I usually like the "-ing" form. Also, from this message I'm not sure why it's important for me (a user) to know this informaiton. If I understand correctly the issue we want the user to be aware is that they'll get plot for a range of years that is different than the standard 5-years span defined by PACTA, right?

If so, here is my suggestion:


```r
span <- diff(range(data$year, na.rm = TRUE))
if (span > 5L) {
  inform(glue("Plotting `technology_share` values for more than the 5 years span defined by PACTA".))
}

if (span < 5L) {
  ... Plotting ... for less than the 5 years defined by PACTA ...
}
```

R/plot_emission_intensity.R
```diff
@@ -32,28 +32,37 @@ check_plot_emission_intensity <- function(data, env = parent.frame()) {
...
-    mutate(year = lubridate::make_date(.data$year))
+prep_emission_intensity <- function(data, convert_label = identity, span_5yr = FALSE) {
+  out <- data %>%
+    prep_common() %>%
+    mutate(label = convert_label(.data$label))
 
-  metrics <- distinct(prep, .data$emission_factor_metric)
+  if (span_5yr) {
+    out <- span_5yr(out)
+  }
```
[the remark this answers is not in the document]
MonikaFu: not entirely I think because the 'emission_intensity' has to be handled differently (because of the use fo `lubridate`, we might want to align with the rest at some point).

R/plot_techmix.R
MonikaFu: See up. I consider it nice to have but not for this release.

R/plot_techmix.R
MonikaFu: Indeed - I inform in both cases as it is not related to the time span but to the way that this function is working at the moment (plotting for extreme years in the dataset). Not sure about all the replacements except for the 'data'. I feel like 'extreme years' might be more useful than the actual years (which they see in the plot anyway). 

Do you mean to extract both `techmix_labels()` and `colours` ?

R/utils.R
MonikaFu: PACTA convention. I guess that ICE is a very well-known abbreviation and the other two not as much.

tests/testthat/_snaps/qplot_emission_intensity.md
MonikaFu: Good question! We need a unit!

tests/testthat/test-plot_emission_intensity.R
MonikaFu: 😆 alright will try to amend

tests/testthat/test-qplot_emission_intensity.R
MonikaFu: Yes, true. In this case it is not the right way to do it

PR #368, opened 2021-06-27
Fix label tech

MonikaFu:
In this PR I fix techmix code so that it uses custom 'label_tech' if present in the data.

(I also moved commonly used functions to `utils.R`.


MonikaFu:
FYI @maurolepore I'm merging this one without review as I am confident that it works and it is actually a fix of a bug.

PR #370, opened 2021-06-28
Modify annotation space

### LINKED ISSUE #347: Clarify ratio of annotation space to plot space 1:4
```r
# R/plot_trajectory.R
    # ASK: Does `6` have a meaning? e.g. `some_space <- 6`. I changed the
    # value to 1-30 and noticed no effect on the plot. Strange.
    xlim = c(min(data$year), max(data$year) + 6)
```

It's a limit so it will not necessarily impact the plot. The 6 is arbitrary and I actually need to change it something that is varying depending on the length of the time interval we are plotting. The point was basically to have a ratio of annotation space to plot space 1:4

_Originally posted by @MonikaFu in https://github.com/2DegreesInvesting/r2dii.plot/pull/339#discussion_r656154845_

MonikaFu:
Closes #347 

In this PR I make the space left for annotations data-dependent instead of hardcoded 6 years. 

Minor change: wrap the trajectory title.

### DIFF
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -104,19 +104,30 @@ plot_trajectory_impl <- function(data) {
   )
 
   lines_end <- filter(order_trajectory(data), .data$year == max(data$year))
-  issue_346 <- 6
+  year_span <- max(data$year) - min(data$year)
   p <- p + ggrepel::geom_text_repel(
     data = lines_end,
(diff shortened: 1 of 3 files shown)


MonikaFu:
FYU @maurolepore - all checks passed and I tested locally that the graphs are generated correctly. I merge without review.

PR #374, opened 2021-06-28
Order techmix bars

### LINKED ISSUE #365: Make sure that plot_techmix plots bars in the right order
Not sure but I think that right now it plots bars as they appear in the data. Should always be: Portfolio, Benchmark, Scenario

MonikaFu:
Closes #365 

In this PR I extend the functionality of `techmix_labels()` function to order the labels.


maurolepore (approved):
LGreatTM.

PR #384, opened 2021-07-28
Fix theme

MonikaFu:
In this PR I pass the `base_size` argument passed by the user to `theme_classic()` so that all the sizes in the plot are adjusted correctly.


maurolepore:
Favoring #387

PR #385, opened 2021-07-28
Use quosures

maurolepore:
This PR replaces the fragile metaprogramming we did with `substitute()` with a more robust metaprogramming with tidy eval. Internally, we now pass quosures, which hold an expression AND
the environment where the expression should be evaluated.

This solves the problem I was facing by which the name of a dataset was lost if the call to
`substitute()` was too deep, e.g. the user passed `bad_data` to `f <- funciton(data) { ... }` 
and they got an error like "Bad `data`." rather than "Bad `bad_data`.".

By solving this problem we can now move the `check_*()` functions into the `prep_*()`
funcitions, thus, allowing us to have a DRYer code base.

Maybe we can meet to discuss why I think this is a good idea. 

Note I first used a different approach (ab8510d): Capture the name of the dataset at the top level then
passing it down to every function that needed it. While that approach is straightforward, it needs a lot
more code than passing a quosure, from which we can pull what we need (i.e. the data or the name of 
the data) wherever we need it.
References:

* https://adv-r.hadley.nz/metaprogramming.html
* http://adv-r.had.co.nz/Computing-on-the-language.html#substitute


maurolepore:
Closing to in favor of #386 which solves the same main problem with a smaller diff.

PR #386, opened 2021-07-29
Use a list as an environment for substitute()

maurolepore:
Replace `env  = parent.frame()` with `env = list(data = substitute(data))`

This is more flexible: It allows to move the make the final
substitution at any depth in the call stack.

This solves the same main concern than #385 with less changes.
My focus here was to make the diffs as small as possible.

--

Reference: http://adv-r.had.co.nz/Computing-on-the-language.html#substitute

### DIFF
--- a/R/plot_techmix.R
+++ b/R/plot_techmix.R
@@ -27,13 +27,11 @@
 #'
 #' plot_techmix(data)
 plot_techmix <- function(data) {
-  check_plot_techmix(data)
-
-  prep <- prep_techmix(data)
+  prep <- prep_techmix(data, env = list(data = substitute(data)))
   plot_techmix_impl(prep)
 }
 
-check_plot_techmix <- function(data, env = parent.frame()) {
+check_plot_techmix <- function(data, env) {
   stopifnot(is.data.frame(data))
   crucial <- c(common_crucial_market_share_columns(), "technology_share")
   hint_if_missing_names(abort_if_missing_names(data, crucial), "market_share")
@@ -76,7 +74,10 @@ abort_if_multiple_scenarios <- function(data, env = parent.frame()) {
 prep_techmix <- function(data,
                          convert_label = identity,
                          span_5yr = FALSE,
-                         convert_tech_label = identity) {
+                         convert_tech_label = identity,
+                         env = NULL) {
+  check_plot_techmix(data, env = env)
+
   out <- data %>%
     prep_common() %>%
     add_label_tech_if_missing() %>%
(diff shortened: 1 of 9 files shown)


R/plot_techmix.R
```diff
@@ -94,7 +96,7 @@ prep_techmix <- function(data,
   start_year <- min(out$year)
   future_year <- max(out$year)
   if (!quiet()) {
-    .data <- deparse_1(substitute(data, env = parent.frame()))
+    .data <- deparse_1(substitute(data, env = env))
```
maurolepore, line 99: ^1
Here we also use `env` and now use the list we create from the top level function rather than the more fragile `parent.frame()` (which breaks if we move this function deeper in the call stack).

tests/testthat/_snaps/plot_techmix.md
```diff
@@ -41,8 +41,8 @@
 # informs that extreme years are used
 
     Code
-      invisible(plot_techmix(data))
+      invisible(plot_techmix(mydata))
     Message <message>
       The `technology_share` values are plotted for extreme years.
-      Do you want to plot different years? E.g. filter data with:`subset(data, year %in% c(2020, 2030))`.
+      Do you want to plot different years? E.g. filter mydata with:`subset(mydata, year %in% c(2020, 2030))`.
```
maurolepore, line 47: Relates to ^1
This snapshot tests that the user actually sees the name given to `data` (here `mydata`) and not the literal "data".

tests/testthat/test-plot_techmix.R
```diff
@@ -91,7 +91,7 @@ test_that("with input data before start year of 'projected' prep_techmix
 })
 
 test_that("informs that extreme years are used", {
-  data <- filter(
+  mydata <- filter(
```
maurolepore, line 94: Relates to ^1

tests/testthat/test-plot_techmix.R
```diff
@@ -101,7 +101,7 @@ test_that("informs that extreme years are used", {
 
   restore <- options(r2dii.plot.quiet = FALSE)
   expect_snapshot(invisible(
-    plot_techmix(data)
+    plot_techmix(mydata)
```
maurolepore, line 104: Relates to ^1

maurolepore (commented):
@MonikaFu I finished moving `check_plot_*()` back to the top-level functions (`plot_*()` and `qplot_*()`). I'm ready for your review.

MonikaFu (approved):
LGTM

PR #387, opened 2021-07-29
Adjust the base_size according to specs

MonikaFu:
In this PR I pass the `base_size` argument passed by the user to `theme_classic()` so that all the sizes in the plot are adjusted correctly.

Makes the #384 redundant.


maurolepore (approved):
LGTM

PR #391, opened 2021-09-21
Catch up with R CMD check workflow in r-lib/actions

maurolepore:
The GH actions we use mostly come from r-lib/actions. This PR updates
the R CMD check workflow to catch up with latest changes in r-lib/actions.

This is not crucial for the workings of this package but I think it's nice to
not let this repo fall behind the improvements in r-lib/actions and the
changes in other r2dii packages.


MonikaFu (approved):
LGTM. Thanks

PR #392, opened 2021-09-21
Reuse _pkgdown.yml from r2dii.analysis

maurolepore:
I see pkgdown.yml is a lil different than that of other r2dii
packages. This PR makes it consistent with other r2dii packages.


MonikaFu (approved):
LGTM

PR #394, opened 2021-09-22
Remove link to r2dii

maurolepore:
The r2dii meta package is poorly maintained so it's likely best
to not point external users to it.


MonikaFu (approved):
LGTM

PR #395, opened 2021-09-24
Dont center trajectory

MonikaFu:
closes ADO [ticket 1907](https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1907/)

In this PR I add a parameter that allows the user to choose whether the trajectory charts are centred around the starting value or not (in the current version of the package they were always centered).

### DIFF
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -242,7 +242,7 @@ scenario_colour <- function(data) {
     arrange(desc(.data$value)) %>%
     pull(.data$metric) %>%
     as.character()
-  num_scen_areas <- length(ordered_scenarios)
+  num_scen_areas <- length(ordered_scenarios) + 1
   scenario_colours <- get_ordered_scenario_colours(num_scen_areas)
 
   technology_kind <- r2dii.data::green_or_brown %>%
@@ -252,11 +252,11 @@ scenario_colour <- function(data) {
 
   switch(technology_kind,
     "green" = reverse_rows(tibble(
-      scenario = ordered_scenarios,
+      scenario = c(ordered_scenarios, c("target_worse")),
       colour = scenario_colours$hex
     )),
     "brown" = tibble(
-      scenario = rev(ordered_scenarios),
+      scenario = rev(c("target_worse", ordered_scenarios)),
       colour = scenario_colours$hex
     ),
     abort( # nocov start
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -1,10 +1,19 @@
 #' Create a trajectory plot
 #'
-#' @param data A data frame. Requirements:
-#'   * The structure must be like [market_share].
-#'   * The following columns must have a single value: `sector`, `technology`,
-#'   `region`, `scenario_source`.
-#'   * (Optional) If present, the column `label` is used for data labels.
+#' @param data A data frame. Requirements: * The structure must be like
+#'   [market_share]. * The following columns must have a single value: `sector`,
+#'   `technology`, `region`, `scenario_source`. * (Optional) If present, the
+#'   column `label` is used for data labels.
+#' @param convert_label A symbol. The unquoted name of a function to apply to
+#'   legend labels. For example, to convert labels to uppercase use
+#'   `convert_label = toupper`. To get the default behaviour of
+#'   `qplot_trajectory()` use `convert_label = format_label`.
+#' @param span_5yr Logical. Use `TRUE` to restrict the time span to 5 years from
+#'   the start year (the default behaviour of `qplot_trajectory()`), or use
+#'   `FALSE` to impose no restriction.
+#' @param center_y Logical. Use `TRUE` to center the y-axis around start value
+#'   (the default behaviour of `qplot_trajectory()`), or use `FALSE` to not
+#'   center.
 #'
 #' @seealso [market_share].
 #'
@@ -22,15 +31,15 @@
 #' )
 #'
 #' plot_trajectory(data)
-plot_trajectory <- function(data) {
+plot_trajectory <- function(data, convert_label = identity, span_5yr = FALSE, center_y = FALSE) {
   env <- list(data = substitute(data))
   check_plot_trajectory(data, env = env)
 
   data %>%
     prep_trajectory(
-      convert_label = identity,
-      span_5yr = FALSE,
-      center_y = FALSE) %>%
+      convert_label,
+      span_5yr,
+      center_y) %>%
     plot_trajectory_impl()
 }
 
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -1,9 +1,10 @@
 #' Create a trajectory plot
 #'
-#' @param data A data frame. Requirements: * The structure must be like
-#'   [market_share]. * The following columns must have a single value: `sector`,
-#'   `technology`, `region`, `scenario_source`. * (Optional) If present, the
-#'   column `label` is used for data labels.
+#' @param data A data frame. Requirements:
+#'   * The structure must be like [market_share].
+#'   * The following columns must have a single value: `sector`, `technology`,
+#'   `region`, `scenario_source`.
+#'   * (Optional) If present, the column `label` is used for data labels.
 #' @param convert_label A symbol. The unquoted name of a function to apply to
 #'   legend labels. For example, to convert labels to uppercase use
 #'   `convert_label = toupper`. To get the default behavior of
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -5,8 +5,6 @@
 #'   * The following columns must have a single value: `sector`, `technology`,
 #'   `region`, `scenario_source`.
 #'   * (Optional) If present, the column `label` is used for data labels.
-#' @param center.y.axis Logical. Use `TRUE` to center the y-axis around start
-#' value, or use `FALSE` to not center.
 #'
 #' @seealso [market_share].
 #'
@@ -24,13 +22,16 @@
 #' )
 #'
 #' plot_trajectory(data)
-plot_trajectory <- function(data, center.y.axis = TRUE) {
+plot_trajectory <- function(data) {
   env <- list(data = substitute(data))
   check_plot_trajectory(data, env = env)
 
   data %>%
-    prep_trajectory(convert_label = identity, span_5yr = FALSE) %>%
-    plot_trajectory_impl(center.y.axis)
+    prep_trajectory(
+      convert_label = identity,
+      span_5yr = FALSE,
+      center_y = FALSE) %>%
+    plot_trajectory_impl()
 }
 
 check_plot_trajectory <- function(data, env) {
@@ -52,10 +53,14 @@ check_plot_trajectory <- function(data, env) {
 #'   uppercase use `convert_label = toupper`.
 #' @param span_5yr Logical. Use `TRUE` to restrict the time span to 5 years from
 #'   the start year, or use `FALSE` to impose no restriction.
+#' @param center_y Logical. Use `TRUE` to center the y-axis around start
+#' value, or use `FALSE` to not center.
+#'
 #' @noRd
 prep_trajectory <- function(data,
                             convert_label = identity,
-                            span_5yr = FALSE) {
+                            span_5yr = FALSE,
+                            center_y = FALSE) {
   out <- data %>%
     prep_common() %>%
     mutate(value = .data$production) %>%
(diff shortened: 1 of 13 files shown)


MonikaFu:
@maurolepore this is still a draft version as I would like to hear your thoughts as to how to implement this feature. At the moment I added a parameter to `plot_trajectory()` that allows the user to switch off the centering. I am not sure if we should add this parameter to `qplot_trajectory()` as well. From theoretical point of view I would think not because it will make the plot less 'quick'. But then from what I've seen the usuers usually use the quick plots so forcing them to use the 'plot' in this case and doing all the data manipulations that quick plot does for them might not be the most-friendly approach. Curious to hear your thoughts.

maurolepore (commented):
Here I share some comments about the implementation. 

I'll think a bit more about how to let the user control this behaviour. I need to refresh how we previously managed to allowed users to control behaviour whithout adding arguments. 

Will be back :)

maurolepore (commented):
Okay, I reviewed the source code and noted we implemented features via arguments to the `prep_*()` functions. The `qplot_*()` functions call `prep_*()` with the argument that displays the most common behaviour and the `plot_*()` functions call `prep_*()` with the argument that displays the most vanilla behaviour. 

For example, `qplot_trajectory()` calls `prep_trajectory()` with `span_5yr = TRUE` and `plot_tajectory()` calls `prep_trajectory()` with `span_5yr = FALSE`. 

If we follow the same approach I think we should add the new argument (I suggest `center_y`) to `prep_tajectory()`, and `qplot_trajectory()` would use `TRUE` while `plot_trajectory()` would use `FALSE`. Does this sound right to you or is there a reason to take a different approach in this particular case?

One road we could take is to expose all arguments passed to `prep_plot()` but maybe only in `plot_trajectory()`. And then document how to use the arguments to match the `qplot_*()` functions. For example, `plot_trajectory()` would gain the argument `span_5yr` and its documentation would look like this:

```r
#' @param span_5yr Logical. Should the plot show only 5 yrs from the start date? 
#'   Use `TRUE` to get the behaviour of `qplot_trajectory()`.
```

That way it should be relatively easy to use `plot_*()` to get a plot like `qplot_*()` except for some feature they prefer turned off.

MonikaFu:
@maurolepore thanks for your comments, it all makes perfect sense. And actually, adding the argument to `prep_trajectory` will simplify significantly `plot_trajectory` I believe. Will work on implementation today.

R/plot_trajectory.R
```diff
@@ -101,7 +101,7 @@ prep_trajectory <- function(data,
   not_scenarios <- out %>%
     filter(!is_scenario(.data$metric)) %>%
     mutate(value_low = .data$value)
-  rbind(scenarios, not_scenarios)
+  bind_rows(scenarios, not_scenarios)
```
maurolepore: The [error we got on CI](https://github.com/2DegreesInvesting/r2dii.plot/runs/3735392598?check_suite_focus=true) suggested the problem is that `dplyr::filter()` got a "matrix" or "list" (instead of a dataframe). Chasing the bug (with ubuntu 20.04 and R 3.6 at https://rstudio.cloud/project/1568984) I noticed that when `rbind()` takes one grouped dataframe it outputs a matrix -- not a dataframe. Replacing `rbind()` with `dplyr::bind_rows()` solved the problem. 

 
```r
no applicable method for 'filter' applied to an object of class "c('matrix', 'list')"
```

It would be nice to replace all instances of `rbind()` with `dplyr::bind_rows()` -- and in general prefer the safer tidyverse alternatives to base functions  (they exist for a reason).

--

This reprex shows that  before R 4.0 `rbind()` was not as smart as `dplyr::bind_rows()`:

``` r
R.version$version.string
#> [1] "R version 3.6.3 (2020-02-29)"

library(dplyr)

x <- data.frame(a = 1)
y <- data.frame(a = 1)

# OK
class(rbind(x, y))
#> [1] "data.frame"

x <- group_by(x, a)

# OK
class(bind_rows(x, y))
#> [1] "grouped_df" "tbl_df"     "tbl"        "data.frame"

# Bad
class(rbind(x, y))
#> [1] "matrix"
```

<sup>Created on 2021-09-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

maurolepore (commented):
@MonikaFu, note the bug is now fixed. You may go ahead with this PR.

R/plot_trajectory.R
MonikaFu: @maurolepore - thank you! I will be using `dplyr::bind_rows` from now on!

MonikaFu:
@maurolepore in commit d2725cf I replaced almost all instances of `rbind` by `bind_rows`. There is one function in `helper.R` in testthat folder still uses `rbind` (line 48 in tests/testthat/helper.R) as I was not sure of its functionality and didn't want to mess anything up. If you think that we should replace it there as well, let me know and I will. Or you go ahead.

R/plot_trajectory.R
```diff
@@ -22,12 +31,15 @@
...
   check_plot_trajectory(data, env = env)
 
   data %>%
-    prep_trajectory(convert_label = identity, span_5yr = FALSE) %>%
+    prep_trajectory(
+      convert_label,
+      span_5yr,
+      center_y) %>%
     plot_trajectory_impl()
 }
```
maurolepore, line 44: If we expose the arguments in the signature of `plot_trajectory()`, then here we need to pass the name of the argument, not its value. If you hard code the value, then you loose the ability to control the behavior from the top-level function:

``` r
# Good
f <- function(x = 1) g(x = x)
g <- function(x = 1) x
f(2)
#> [1] 2

# Bad
f <- function(x = 1) g(x = 1)
g <- function(x = 1) x
f(2)
#> [1] 1
```

<sup>Created on 2021-09-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

--

Maybe write  a test to confirm the top-level function `plot_trajectory()` is indeed sensitive to changing the values of the arguments passed to `prep_trajectory()`.

R/plot_trajectory.R
```diff
@@ -50,10 +62,14 @@ check_plot_trajectory <- function(data, env) {
 #'   uppercase use `convert_label = toupper`.
 #' @param span_5yr Logical. Use `TRUE` to restrict the time span to 5 years from
 #'   the start year, or use `FALSE` to impose no restriction.
+#' @param center_y Logical. Use `TRUE` to center the y-axis around start
+#' value, or use `FALSE` to not center.
+#'
 #' @noRd
```
maurolepore, line 68: If the arguments are documented in an exported function (`plot_trajectory()`), then it's best to not document them in the internal function (`prep_trajectory()`) -- that way we don't need to maintain the docs in two places.

R/plot_trajectory.R
```diff
@@ -198,64 +222,95 @@ order_trajectory <- function(data) {
...
+      )
+  }
+  data
 }
 
-distance_from_start_value <- function(data, value) {
+start_value_portfolio <- function(data) {
   start_value_portfolio <- data %>%
-    filter(.data$year == min(.data$year), is_portfolio(.data$metric)) %>%
+    filter(.data$year == min(data$year), is_portfolio(.data$metric)) %>%
```
maurolepore, line 253: `min()` defaults to `na.rm = FALSE`. What happens when the data has `NA`s? Do we have a test for it? 

https://github.com/2DegreesInvesting/r2dii.plot/issues/326

R/plot_trajectory.R
```diff
@@ -79,18 +95,27 @@ prep_trajectory <- function(data,
...
-    data = scenario(data),
+    data = data %>% filter(is_scenario(metric)),
     aes(
       ymin = .data$value_low,
       ymax = .data$value,
-      fill = .data$metric,
+      fill = factor(
+        .data$metric,
+        levels = scenario_colour(data %>% filter(is_scenario(metric)))$scenario
+        ),
```
maurolepore, line 118: Nit: I would either use the pipe or nested calls -- not a mix of both.

```r
# Pipe
f() %>% g() %>% h()
# Nest
h(g(f()))
# Mix
h(f() %>% g())
```

tests/testthat/test-plot_trajectory.R
```diff
@@ -44,3 +44,17 @@ test_that("the errors message includes the name of the user's data", {
   bad_region$region <- c("a", "b")
   expect_error(plot_trajectory(bad_region), "bad_region")
 })
+
+test_that("By default doesn't center the Y axis", {
```
maurolepore, line 48: For each new argument that we export I think we need to test the top-level function is sensitive to changes in that argument. 



``` r
library(testthat, warn.conflicts = FALSE)


# NEWS.md
# * `f()` gains the argument `x`.

# R/f.R
f <- function(x = TRUE) x

# tests/testthat/test-f.R
test_that("is sensitive to `x", {
  expect_true(f())
  expect_false(f(FALSE))
})
#> Test passed 🎊
```

<sup>Created on 2021-09-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

tests/testthat/test-plot_trajectory.R
```diff
@@ -44,3 +44,17 @@ test_that("the errors message includes the name of the user's data", {
...
+    prep_trajectory(convert_label = identity, span_5yr = FALSE, center_y = FALSE)
+  start_val <- start_value_portfolio(data_prep)
+
+  p <- plot_trajectory(data)
+
+   expect_false(
+    abs(start_val - ggplot_build(p)$layout$panel_scales_y[[1]]$range$range[1]) ==
+    abs(start_val - ggplot_build(p)$layout$panel_scales_y[[1]]$range$range[2])
+    )
+})
```
maurolepore, line 60: It's hard to see that this code is doing what the test-title promises. Maybe use some intermediate variables to give meaningful names to pieces of code that capure relevant concepts such as `centered`, `not_centered`.

The idea is to try help the reader run the code in their heads. To better understand what's going on, they can always run the code but they should not have to.

maurolepore (changes requested):
Thanks Monika!

- I hit request changes because I think there is  a small bug: Inside `plot_trajectory()` we need to pass the promise (not the value) to `prep_trajectory()` -- else the user won't be able to control the output from the top-level function. See my comment below.

Regarding our intention to export new arguments:

- I think we need to increment in the minor component of the version number.
- I think we need to engage Alex so he's aware of this significant change to the public interface.
- I think to we need to consider doing the same for all the `plot_*()` function, for consistency.

MonikaFu:
@maurolepore - I am not sure what you mean by 'passing a promise'. Could you explain more?

R/plot_trajectory.R
```diff
@@ -22,12 +31,15 @@
 #' )
 #'
 #' plot_trajectory(data)
-plot_trajectory <- function(data) {
+plot_trajectory <- function(data, convert_label = identity, span_5yr = FALSE, center_y = FALSE) {
```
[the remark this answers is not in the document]
MonikaFu: yes, your suggestion makes sense to do it for all other functions. I will put it in another issue and work on it separately, though, as I would like like to close this PR as Antoine needs this feature.

maurolepore:
Sorry Monika, I thought I had added a clarifying reprex but I didn't. Comming soon ...

maurolepore:
``` r
# Good:
# f_exported() passes to g_internal() the name that holds the value (promise)
f_exported <- function(x = 1) g_internal(x = x)  # Note x = x not x = 1
g_internal <- function(x = 1) x

# f_exported() is sensitive to changes in `x`
f_exported(1)
#> [1] 1
f_exported(2)
#> [1] 2
f_exported(3)
#> [1] 3



# Bad: 
# f_exported() passes to g_internal() a hard-coded value `1`
f_exported <- function(x = 1) g_internal(x = 1)
g_internal <- function(x = 1) x

# f_exported() is insensitive to changes in `x` -- it always returns `1`
f_exported(1)
#> [1] 1
f_exported(2)
#> [1] 1
f_exported(3)
#> [1] 1
```

<sup>Created on 2021-09-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

MonikaFu:
@maurolepore Thanks for the example! I believe that my implementation was having the desired behaviour but now I made it more explicit.

maurolepore (approved):
LGTM. 

Thanks Monika, I see you addressed my main concern and are aware of my other suggestions. Go ahead and merge when you are ready.

R/plot_trajectory.R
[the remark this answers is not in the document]
MonikaFu: See #398

PR #396, opened 2021-09-25
Use dev version

maurolepore:
This should have been the first commit after
releasing r2dii.plot 0.1.0﻿.

It seems I forgot to do this step -- as it appears
unchecked at https://github.com/2DegreesInvesting/r2dii.plot/issues/369


MonikaFu (approved):
LGTM

PR #397, opened 2021-09-25
Restore Get started

maurolepore:
The section "Get started" is gone. I must have removed it by
accident. This commit restores it.


MonikaFu (approved):
LGTM

PR #400, opened 2021-10-07
Remove lubridate dependency

AlexAxthelm:
This PR is pretty simple, in that it changes the `lubridate` functions (of which there were not many) back to `base` functions, so that we can reduce the number of dependencies.

Initially inspired by concerns that `lubridate` would be archived on CRAN (no longer a serious concern), but continued because the fixes are small, and removing dependencies  is generally a good thing.

Type signature of functions is unchanged.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -38,7 +38,6 @@ Imports:
     ggplot2,
     ggrepel,
     glue,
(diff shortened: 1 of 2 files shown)


MonikaFu (commented):
LGTM

PR #399, opened 2021-10-06
Add colour scales

MonikaFu:
Closes #28 

In this PR I add the colour scales that allow the user to easily apply 2DII colour palettes in their plots. Scale functions for 3 colour palettes are added:
* 2DII palette
* Sector palette
* Technology palette

Scale functions for each palette are implemented in a separate file. Their structure is very similar with some differences with regards to the number and names of parameters reflecting the corresponding colour datasets. 

Also NEWS and articles were updated to inform the users about the new functions.

PR #406, opened 2021-10-08
DRY functions that abort if value is unknown

maurolepore:
This PR extends #399 (`add-colour-scales` -- it merges into it).

Here I remove duplication in functions that abort if
a value passed by the user is unknown. 

I reworded the error messages so that the same function
can be used in multiple situations. The message is a
bit less precise but I think it's still clear enough
and makes the code more maintainable.

### DIFF
--- a/R/r2dii_pal_impl.R
+++ b/R/r2dii_pal_impl.R
@@ -1,6 +1,6 @@
-r2dii_pal_impl <- function(thing, data, column) {
-  thing <- thing %||% data[[column]]
-  values <- as_tibble_col(thing, column) %>%
+r2dii_pal_impl <- function(x, data, column) {
+  x <- x %||% data[[column]]
+  values <- as_tibble_col(x, column) %>%
     inner_join(data, by = column) %>%
     pull(.data$hex)
   max_n <- length(values)


R/imports.R
```diff
@@ -17,3 +17,8 @@ utils::globalVariables(c(
   "where",
   ":="
 ))
+
+# Re-implement tibble::as_tibble_col to avoid adding tibble to Imports
+as_tibble_col <- function (x, column_name = "value") {
+  tibble(`:=`(!!column_name, x))
+}
```
maurolepore, line 24: I didn't realize we were not using tibble in Imports.

R/r2dii_pal_impl.R
```diff
@@ -1,6 +1,6 @@
-r2dii_pal_impl <- function(thing, data, column) {
-  thing <- thing %||% data[[column]]
-  values <- tibble::as_tibble_col(thing, column) %>%
+r2dii_pal_impl <- function(x, data, column) {
+  x <- x %||% data[[column]]
+  values <- as_tibble_col(x, column) %>%
```
maurolepore, line 3: `x` is more common than `thing`, shorter, and equally vague.

PR #408, opened 2021-10-08
DRY `r2dii_*_pal()`

maurolepore:
Extends #399 

This PR removes duplication in `r2dii_*pal()` functions. I write the implementation in a separate file to minimize merge conflicts. Later we can move it to R/utils.R or wherever feels appropriate.

I'm relying on tests but have not inspected the output visually. Monika, can see if the output is still as expected?


MonikaFu (approved):
LGTM. Although I am not sure if 'thing' is the best name here. Maybe a bit too general. But leaving it for now.

PR #410, opened 2021-10-09
Rename files after main topic

maurolepore:
Extends #399 

This PR renames files in R and tests/testthat so they match the main topic. This plays well with `use_r()` and `use_test()`.


MonikaFu (approved):
LGTM although I see some merge conflicts that need to be resolved.

PR #413, opened 2021-10-12
Remove ggplot2 arguments 'data' and 'mapping'

maurolepore:
Extends #399 

Follow up #411

> I agree with removing 'data' and 'mapping'
-- https://github.com/2DegreesInvesting/r2dii.plot/pull/411#pullrequestreview-777124776


MonikaFu (changes requested):
I merged #399 into `develop` and it seems to be causing some issues with the history :/ I am sorry for that. I guess that we need to cherry-pick the last commit from this one and make another PR?

PR #421, opened 2021-10-13
(Extend #417) DRY parameters via templates and inheritance

maurolepore:


man-roxygen/convert_label.R
```diff
@@ -0,0 +1,5 @@
+#' @param convert_label A symbol. The unquoted name of a function to apply to
+#'   y-axis labels. For example:
+#'   * To convert labels to uppercase use `convert_label = toupper`.
+#'   * To get the default behavior of `<%= fun%>()` use
+#'   `convert_label = <%= value%>`.
```
maurolepore, line 5: @jdhoffa I learned a new trick you may like.

jdhoffa: I do indeed

PR #422, opened 2021-10-13
(Extend #417) DRY tests

maurolepore:
- DRY tests with a lil helper

### DIFF
--- a/tests/testthat/test-plot_emission_intensity.R
+++ b/tests/testthat/test-plot_emission_intensity.R
@@ -61,12 +61,8 @@ test_that("is sensitive to `span_5yr`", {
   data <- filter(sda, sector == "cement")
 
   p_f <- plot_emission_intensity(data, span_5yr = FALSE)
-  min_year <- min(as.numeric(format(p_f$data$year, format = "%Y")), na.rm = TRUE)
-  max_year <- max(as.numeric(format(p_f$data$year, format = "%Y")), na.rm = TRUE)
-  expect_false(max_year - min_year == 5)
+  expect_false(diff(year_range(p_f)) == 5)
 
   p_t <- plot_emission_intensity(data, span_5yr = TRUE)
-  min_year <- min(as.numeric(format(p_t$data$year, format = "%Y")), na.rm = TRUE)
-  max_year <- max(as.numeric(format(p_t$data$year, format = "%Y")), na.rm = TRUE)
-  expect_true(max_year - min_year == 5)
+  expect_true(diff(year_range(p_t)) == 5)
 })
--- a/tests/testthat/test-plot_emission_intensity.R
+++ b/tests/testthat/test-plot_emission_intensity.R
@@ -54,7 +54,7 @@ test_that("is sensitive to `convert_label`", {
   labels_mod <- plot_emission_intensity(data, convert_label = toupper) %>%
     unique_plot_data("label")
 
-  expect_true(difference_is_letter_case(labels_def, labels_mod))
+  expect_false(identical(labels_def, labels_mod))
 })
 
 test_that("is sensitive to `span_5yr`", {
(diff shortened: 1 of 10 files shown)


tests/testthat/test-scale_colour_r2dii.R
```diff
@@ -13,26 +13,20 @@ test_that("if with bad `labels` errors gracefully", {
...
-  colours_default <- unique(g$data[[1]]["colour"])
 
-  p_after <- p + scale_colour_r2dii()
-  g_after <- ggplot_build(p_after)
-  colours_changed <- unique(g_after$data[[1]]["colour"])
+  colours_default <- unique_data1(p, "colour")
+  colours_changed <-  unique_data1(p + scale_colour_r2dii(), "colour")
 
-  expect_false(isTRUE(all.equal(colours_default, colours_changed)))
+  expect_false(identical(colours_default, colours_changed))
```
maurolepore, line 20: Monika, did I interpret this expectation correctly?

tests/testthat/test-plot_techmix.R
```diff
@@ -267,13 +260,10 @@ test_that("is sensitive to `convert_tech_label`", {
...
-  p_mod <- plot_techmix(data, convert_tech_label = toupper)
-  g_mod <- ggplot_build(p_mod)
-  labels_mod <- unique(g_mod$plot$data$label_tech)
+  labels_def <- plot_techmix(data) %>%
+    unique_plot_data("label_tech")
+  labels_mod <- plot_techmix(data, convert_tech_label = toupper) %>%
+    unique_plot_data("label_tech")
 
-  expect_false(isTRUE(all.equal(labels_def, labels_mod)))
+  expect_true(difference_is_letter_case(labels_def, labels_mod))
```
maurolepore, line 268: Monika, did I interpret this expectation correctly?

MonikaFu: no, not necessarily, although in this case it will be. The point of this test is basically to check whether after passing a function modifying labels to `convert_label` paratemer we indeed get different labels. But whether the difference is letter case is not the point. If we passed a different function then the difference would be somethig else. So it is a wrong test in my opinion.

tests/testthat/test-plot_emission_intensity.R
```diff
@@ -49,27 +49,20 @@ test_that("with too many lines to plot errors gracefully", {
...
-  p_mod <- plot_emission_intensity(data, convert_label = toupper)
-  g_mod <- ggplot_build(p_mod)
-  labels_mod <- unique(g_mod$plot$data$label)
+  labels_def <- plot_emission_intensity(data) %>%
+    unique_plot_data("label")
+  labels_mod <- plot_emission_intensity(data, convert_label = toupper) %>%
+    unique_plot_data("label")
 
-  expect_false(isTRUE(all.equal(labels_def, labels_mod)))
+  expect_true(difference_is_letter_case(labels_def, labels_mod))
```
MonikaFu, line 57: This is not really correct. What I am testing for here is whether the labels differ from each other. I feel like formulating it this was draws attention away from what the test is acually about. But maybe I am wrong.

tests/testthat/test-plot_trajectory.R
```diff
@@ -81,27 +81,20 @@ test_that("Is sensitive to `center_y`", {
...
 })
 
 test_that("is sensitive to `span_5yr`", {
   data <- example_market_share()
 
   p_f <- plot_trajectory(data, span_5yr = FALSE)
-  min_year <- min(p_f$data$year, na.rm = TRUE)
-  max_year <- max(p_f$data$year, na.rm = TRUE)
-  expect_false(max_year - min_year == 5)
+  expect_false(diff(year_range(p_f)) == 5)
```
MonikaFu, line 96: neat

MonikaFu (changes requested):

MonikaFu (commented):
LGTM. Thank you!

PR #423, opened 2021-10-13
(Extend #417) In article, wrap lines and style

maurolepore:
- Wrap article at 80 char for nicer diffs
- Run styler


maurolepore:
Here is a better example:

https://gist.github.com/maurolepore/e482dc12b113c9da5c7ff7c061b53e76/revisions

PR #425, opened 2021-10-14
Reword

maurolepore:
This PR should fail to render the article Get started because we use but don't export `to_title()` and friends. First merge #426.


vignettes/r2dii-plot.Rmd
```diff
@@ -230,14 +224,13 @@ plot_techmix(data) +
...
+A number of functions allow you to customize your plots using the 2DII style:
 
-* Use `theme_2dii()` to display non-data content in 2DII style.
+* Use `theme_2dii()` to display non-data content.
 
 ```{r}
-# plot change in production plans over the next 5 years for technologies in
-# 'power' sector
+# focus on production plans for technologies in the 'power', globally, for 5
+# years since 2020
```
maurolepore, line 233: I would remove this and all comments because they focus the readers attention on the subset of data rather than the plot.

MonikaFu: I'm fine with that. I thought that it might be a good hint for the users for what they could plot with the analysis output but you are right that it might draw the attention to wrong things.

PR #437, opened 2021-10-15
Polish NEWS

MonikaFu:


NEWS.md
```diff
@@ -8,9 +10,6 @@
     - `recode_metric_techmix()`
     - `spell_out_technology()`
 
-* `plot_trajectory()` now defaults to not center the y-axis around the
-start value.
-
 * `plot_*()` functions gains new parameters to help replicate some of
```
maurolepore, line 13: While you're here, can you change "gains" to "gain"? It's plural so the "s" doesn't apply.

NEWS.md
```diff
@@ -23,6 +22,22 @@ start value.
...
+* `plot_trajectory()` now displays benchmark trajectory lines with improved 
+visibility (#389 @Antoine-Lalechere).
+
+Bug fixes
+
+* `plot_trajectory()` now displays only integer years on the x-axis (#403).
+
+* The 'Get started' section of the website works again.
+
+* `theme_2dii()` correctly uses the 'base_size' argument now.
```
maurolepore, line 40: Maybe reads better as "... now correctly uses ..."

PR #457, opened 2022-02-11
Prove that CI fails with an empty commit

maurolepore:


maurolepore:
@MonikaFu note snapshots fail on `develop` (this PR adds an empty commit). I pushed a PR updating the snapshots (#458).

PR #462, opened 2022-04-04
Update to align with changes in r2dii.analysis v 0.1.13

MonikaFu:
In this PR I am updating the package to align it with the recent changes in `r2dii.analysis`. In particular:

* update test datasets `sda` and `market_share`
* use new columns from `market_share` in volume trajectory plots (new parameter `value_col` specifies which column to plot)
* change the y-axis scale to a percentage (controlled by a parameter `perc_y_scale`)
* update quick plot defaults and labels

**Note:** I'm trying to set up a GH action to check if these changes are compatible with devel versions of other `r2dii` packages (#466). The full check won't work as the changes are not yet on CRAN.


MonikaFu:
Currently, the check is failing as the `r2dii.analysis` version it is aligned with is not on CRAN.

R/plot_trajectory.R
```diff
@@ -1,10 +1,9 @@
...
 #'
-#' @param data A data frame. Requirements:
-#'   * The structure must be like [market_share].
-#'   * The following columns must have a single value: `sector`, `technology`,
-#'   `region`, `scenario_source`.
-#'   * (Optional) If present, the column `label` is used for data labels.
+#' @param data A data frame. Requirements: * The structure must be like
+#'   [market_share]. * The following columns must have a single value: `sector`,
+#'   `technology`, `region`, `scenario_source`. * (Optional) If present, the
+#'   column `label` is used for data labels.
```
jdhoffa, line 6: Is this formatted correctly?

R/plot_trajectory.R
```diff
@@ -156,19 +151,33 @@ plot_trajectory_impl <- function(data) {
...
     theme(plot.margin = unit(c(0.5, 4, 0.5, 0.5), "cm"))
 }
 
+check_y_scale_parameter <- function(value) {
+  if (!is.logical(value)) {
+    abort(c("`perc_y_scale` must be of type logical.",
+            x = glue("You passed a {typeof(value)}.")))
+  }
+}
+
```
jdhoffa, line 180: NIT: Usually for checking input parameters, I just write something like this at the top of the function (and check all of them at once): 
``` r
  stopifnot(
    is.data.frame(data),
    is.data.frame(ald),
    is.data.frame(scenario),
    is.data.frame(region_isos),
    is.logical(use_credit_limit),
    is.logical(by_company),
    is.logical(weight_production)
  ) 
```

rather than writing a custom function to check the parameter. Def not saying you need to do this too, but just FYI :-)

R/qplot_trajectory.R
```diff
@@ -25,31 +25,40 @@
...
 
   data %>%
-    prep_trajectory(convert_label = format_metric, span_5yr = TRUE, center_y = TRUE) %>%
-    plot_trajectory_impl() %>%
-    labs_trajectory()
+    prep_trajectory(
+      convert_label = format_metric,
+      span_5yr = TRUE,
+      center_y = TRUE,
+      value_col = "percentage_of_initial_production_by_scope") %>%
```
jdhoffa, line 38: style: if you're following tidy style, the last bracket should be on the following line

R/qplot_trajectory.R
```diff
@@ -25,31 +25,40 @@
...
+  technology <- spell_out_technology(p[["data"]][["technology"]][[1]])
   sector <- tools::toTitleCase(p[["data"]][["sector"]][[1]])
   min_year <- min(p[["data"]][["year"]], na.rm = TRUE)
+  scope <- data$scope[1]
 
   p +
     labs(
       title = glue(
-        "Production Trajectory of Technology: {tech}
+        "Production Trajectory of {technology} Technology
```
jdhoffa, line 52: These changes seem unrelated to this PR. More focused on changing how the function handles `technology` input.

jdhoffa, line 52: These changes seem unrelated to this PR. More focused on changing how the function handles `technology` input.

tests/testthat/test-plot_techmix.R
```diff
@@ -145,7 +145,7 @@ test_that("Does not output pretty labels", {
   p <- plot_techmix(data)
 
   metrics <- unique(p$data$label)
-  ugly <- c("projected", "corporate_economy", "target_sds")
+  ugly <- c("projected", "target_sds", "corporate_economy")
```
jdhoffa, line 148: If you use:
`expect_equal(sort(vector1), sort(vector2))`
you won't ned to worry about the order of the entries being the same

jdhoffa (commented):
A couple of NIT comments here. 

My main comment is that there are a few changes in this PR that aren't really related to this PR (ie. the updates to the `sda` dataset are sorta separate). 

As the maintainer, I'll leave it up to you if you want to split these changes into two PRs to keep a tidy git history, or if you just want to merge them all at once.

MonikaFu:
Thank you @jdhoffa ! I will keep all the changes in this PR, even though I realise that some of them could more correctly be put in different PRs. If I assume that this PR is about aligning with the changes in `r2dii.analysis` then it makes sense to me to, for example, update both `sda` and `market_share` datasets. As the update was an iterated process, the commits became somehow convoluted and I don't want to go into the mess of disentangling them.

R/qplot_trajectory.R
MonikaFu: Thanks. In principle, I agree with you. However, it is such a small change, and it relates only to labels which I was changing anyway because of the `r2dii.analysis` update that I will keep it in.

R/plot_trajectory.R
MonikaFu: No, it isn't. Thanks for spotting. Reflowing the comment destroyed the list structure.

R/plot_trajectory.R
```diff
@@ -1,10 +1,10 @@
...
 #'
 #' @param data A data frame. Requirements:
-#'   * The structure must be like [market_share].
-#'   * The following columns must have a single value: `sector`, `technology`,
-#'   `region`, `scenario_source`.
-#'   * (Optional) If present, the column `label` is used for data labels.
+#' * The structure must be like [market_share].
+#' * The following columns must have a single value: `sector`, `technology`,
+#' `region`, `scenario_source`.
+#' * (Optional) If present, the column `label` is used for data labels.
```
jdhoffa, line 7: I think these need to be indented? Have you checked that the documentation renders properly

R/plot_trajectory.R
```diff
@@ -14,6 +14,11 @@
 #' @param center_y Logical. Use `TRUE` to center the y-axis around start value
 #'   (the default behavior of `qplot_trajectory()`), or use `FALSE` to not
 #'   center.
+#' @param value_col Character. Name of the column to be used as a value to be
```
jdhoffa, line 17: Would also mention default behaviour here

jdhoffa (approved):
Some minor comments, otherwise lgtm

R/plot_trajectory.R
MonikaFu: Following @maurolepore advice from a while back I don't mention the defaults of the parameters because it is visible right away in the usage section of the documentation:

image

jdhoffa: I see! Fair enough

PR #466, opened 2022-04-20
Add action checking compatibility with development versions of r2dii

MonikaFu:
In this PR I add a GitHub action to check if the package works with devel versions of other `r2dii` packages.

I based it on GH action from `r2dii.analysis` package: https://github.com/2DegreesInvesting/r2dii.analysis/blob/main/.github/workflows/check-full_r2dii-devel.yaml

### DIFF
--- a/.github/workflows/check-full.yaml
+++ b/.github/workflows/check-full.yaml
@@ -56,15 +56,3 @@ jobs:
           extra-packages: rcmdcheck
 
       - uses: r-lib/actions/check-r-package@v1
-
-      - name: Show testthat output
-        if: always()
-        run: find check -name 'testthat.Rout*' -exec cat '{}' \; || true
-        shell: bash
-
-      - name: Upload check results
-        if: failure()
-        uses: actions/upload-artifact@main
-        with:
-          name: ${{ runner.os }}-r${{ matrix.config.r }}-results
-          path: check
--- a/.github/workflows/check-full.yaml
+++ b/.github/workflows/check-full.yaml
@@ -6,9 +6,9 @@
 # usethis::use_github_action("check-standard") will install it.
 on:
   push:
-    branches: [main, master]
+    branches: [develop, main, master]
   pull_request:
-    branches: [main, master]
+    branches: [develop, main, master]
 
 name: R-CMD-check
 
--- a/.github/workflows/check-full.yaml
+++ b/.github/workflows/check-full.yaml
@@ -6,9 +6,9 @@
 # usethis::use_github_action("check-standard") will install it.
 on:
   push:
-    branches: [develop, main, master]
+    branches: [main, master]
   pull_request:
-    branches: [develop, main, master]
+    branches: [main, master]
 
 name: R-CMD-check
 
--- a/.github/workflows/check-full_r2dii-devel.yaml
+++ b/.github/workflows/check-full_r2dii-devel.yaml
@@ -63,15 +63,3 @@ jobs:
         shell: Rscript {0}
 
       - uses: r-lib/actions/check-r-package@v1
-
-      - name: Show testthat output
-        if: always()
-        run: find check -name 'testthat.Rout*' -exec cat '{}' \; || true
-        shell: bash
-
-      - name: Upload check results
-        if: failure()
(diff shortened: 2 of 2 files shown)


.github/workflows/check-full_r2dii-devel.yaml
```diff
@@ -0,0 +1,77 @@
...
+        if: always()
+        run: find check -name 'testthat.Rout*' -exec cat '{}' \; || true
+        shell: bash
+
+      - name: Upload check results
+        if: failure()
+        uses: actions/upload-artifact@main
+        with:
+          name: ${{ runner.os }}-r${{ matrix.config.r }}-results
+          path: check
```
maurolepore, line 77: This whole chunk seems to have been removed from the [original action](https://github.com/r-lib/actions/blob/v2-branch/examples/check-full.yaml).

.github/workflows/check-full_r2dii-devel.yaml
```diff
@@ -0,0 +1,77 @@
...
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+#
+# NOTE: This workflow is overkill for most R packages and
+# check-standard.yaml is likely a better choice.
+# usethis::use_github_action("check-standard") will install it.
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
```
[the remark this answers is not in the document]
MonikaFu: thanks @maurolepore I updated it.

.github/workflows/check-full_r2dii-devel.yaml
MonikaFu: Thanks @maurolepore I aligned with `usethis::use_github_action("check-full")`

.github/workflows/check-full_r2dii-devel.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - uses: r-lib/actions/setup-r-dependencies@v1
+        with:
+          extra-packages: rcmdcheck
+
+      - name: Install devel version of other r2dii packages
+        run: |
+          pak::pkg_install("2DegreesInvesting/r2dii.data")
+          pak::pkg_install("2DegreesInvesting/r2dii.match")
+          pak::pkg_install("2DegreesInvesting/r2dii.analysis")
+        shell: Rscript {0}
```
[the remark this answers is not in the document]
MonikaFu: Thank you for your comments @maurolepore ! Looking at the fact that the checks now fail on R-CDM-check r2dii-devel I think that the action works as expected. They fail because of the output of `r2dii.analysis` changing and those changes not being reflected in tests (this is addressed in PR #462). 

My understanding was that both the full check and the r2dii-devel check will be running now on each PR but this is not the case. I only see the results of devel check in this PR. Do you know why this is the case?

.github/workflows/check-full.yaml
```diff
@@ -6,9 +6,9 @@
 # usethis::use_github_action("check-standard") will install it.
 on:
   push:
-    branches: [develop, main, master]
+    branches: [main, master]
   pull_request:
-    branches: [develop, main, master]
+    branches: [main, master]
```
maurolepore, line 11: I think the reason why this PR doesn't trigger the workflow check-full.yaml is because we need to add the branch `develop` here. 

I'l reopen a can of worms: The fact that r2dii.plot uses `develop` adds little frictions that -- to me -- add more trouble than it is worth.

jdhoffa: Agree

MonikaFu: Indeed. Thank you @maurolepore . I agree with you that maybe `develop` is overkill but I don't have the capacity now to deal with this can of worms unfortunately 😄

MonikaFu:
Looking at the results of the checks. The action works as expected to me. @jdhoffa @AlexAxthelm @maurolepore do you approve?

AlexAxthelm (approved):

jdhoffa (approved):
LGTM, assuming you expect the r2dii-devel checks to fail?

MonikaFu:
@jdhoffa yes, thanks!

Issue #464, opened 2022-04-20
Fix trajectory charts


MonikaFu:
With the updated `market_share` test dataset the lower plot area boundary is cut in an unexpected way. 

``` r
library(r2dii.plot)
library(dplyr)
#> Warning: package 'dplyr' was built under R version 4.0.5
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

data <- market_share %>%
  filter(
    sector == "power",
    technology == "oilcap",
    region == "global",
    scenario_source == "demo_2020"
  )

qplot_trajectory(data)
```

![](https://i.imgur.com/WEtLcp4.png)

<sup>Created on 2022-04-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

PR #473, opened 2022-04-26
fix bug trajectory clean

MonikaFu:
Closes https://github.com/2DegreesInvesting/r2dii.plot/issues/464

In this PR I change the way that trajectory chart area borders are calculated so that the 0 values in the data result in a correct plot.


jdhoffa (approved):
lgtm :-)

PR #476, opened 2022-04-28
Label scenario

### LINKED ISSUE #230: In plot_techmix*() we don't konw which scenario we are plotting.
Do we need a title showing which scenario (e.g. "sds") we are plotting?

MonikaFu:
Closes #230 

In this PR I change the labelling functions so that the scenario name is part of `qplot_techmix()` label.


R/r2dii.plot-package.R
```diff
@@ -6,5 +6,6 @@
 ## usethis namespace: start
 #' @importFrom dplyr n
 #' @importFrom scales percent
+#' @importFrom stringr str_count
```
jdhoffa, line 9: I would tend to prefer calling `stringr::str_count` so it's explicit where the `stringr` dependency is?

jdhoffa: Ah it looks like you do anyways. So you can probably remove this line

MonikaFu: I think that if I remove this line then the checks will not pass?

jdhoffa (approved):
Super minor comment, but lgtm

PR #479, opened 2022-04-29
Change label trajectory

### LINKED ISSUE #478: Change portfolio label qplot_trajectory
ADO #3582 :

> I think the default setting for the portfolio line should be called "portfolio projected" rather than just projected. 

> If possible please change this in r2dii.plot and on the website

MonikaFu:
Closes #478 

In this PR I change the way `qplot_trajectory()` is labelled.


R/to_title.R
```diff
@@ -46,16 +46,21 @@ capitalize_single_letters <- function(words) {
...
-recode_metric_techmix <- function(x) {
-  out <- recode_metric(x)
+recode_metric_trajectory <- function(x) {
+  out <- case_when(
+    x == "projected" ~ "portfolio",
+    startsWith(x, "target") ~ recode_scenario(x),
+    TRUE ~ x
+  )
+  out <- sub("scenario", "", out)
+  out <- trimws(out)
```
jdhoffa, line 63: NIT: Just a comment on naming/ refactoring, it seems a weird to have two functions `record_metric_techmix`, and `recode_metric_trajectory`, and only one uses the function `recode_metric`, while the other uses the function `recode_scenario. `

MonikaFu: Thanks! Good thinking. I addressed it in [4b02ed6](https://github.com/2DegreesInvesting/r2dii.plot/pull/479/commits/4b02ed69f915e645d159c0d4cabd8273f2e7c102)

PR #492, opened 2022-09-21
Release/0.0.1

jdhoffa:
There seems to be some funkiness happening with git conflicts in the `master` and `develop` branches. 

I think there was a `release` branch that had commits/ hotfixes on it, that was merged into `master` but not back-merged into `develop`, and so the histories are out of sync. 

I am going to try some successive releases on a new `main` branch (just to test if squash merging the commits between each release from `develop` works as expected). This shouldn't affect anything as main will still contain the full release and tag history of the package.


MonikaFu:
Do I understand correctly that you are re-creating the history of this package by replacing `master` by `main` with a healthy history?

jdhoffa:
> Do I understand correctly that you are re-creating the history of this package by replacing `master` by `main` with a healthy history?

Yes indeed. And I will do so sequentially by re-merging every release from develop into `main` (starting with `0.0.1`)

jdhoffa:
You'll notice that the final commit in this (and subsequent PRs like this) will be the tagged commit on `develop`

Screen Shot 2022-09-21 at 1 50 57 PM

MonikaFu (approved):
LGTM as far as I understand :)

PR #508, opened 2023-05-12
Fix checks

### LINKED ISSUE #500: @jdhoffa yes, sounds good to me. Will you make an issue for it?
These seem to be a result of changes in `r2dii.analysis` and `testthat`.

See here for more context:

_Originally posted by @MonikaFu in https://github.com/RMI-PACTA/r2dii.plot/issues/499#issuecomment-1342498583_

### LINKED ISSUE #507: Review and handle failing tests
This package could benefit from some overall maintenance. There are several failing and warning tests (some are likely quite easy to fix).

MonikaFu:
Closes #500 
Closes #507 

Update code so checks are passing.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -81,11 +81,12 @@ abort_if_missing_names <- function(data, expected_names) {
 
   if (!all(unique(expected_names) %in% names(data))) {
     missing_names <- sort(setdiff(expected_names, names(data)))
-    abort(c(
-      "`data` must have all the expected names.",
-      x = glue("Missing names: {toString(missing_names)}.")
-    ),
-    class = "missing_names"
+    abort(
+      c(
+        "`data` must have all the expected names.",
+        x = glue("Missing names: {toString(missing_names)}.")
+      ),
+      class = "missing_names"
     )
   }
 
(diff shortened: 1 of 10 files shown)


R/utils.R
```diff
@@ -38,7 +38,7 @@ abort_if_has_zero_rows <- function(data, env) {
 }
 
 hint_if_missing_names <- function(expr, like) {
-  rlang::with_handlers(
+  withCallingHandlers(
```
jdhoffa, line 41: N.B. Why do we need calling handlers actually?

jdhoffa (approved):
yay LGTM

R/utils.R
MonikaFu: I believe that the handlers are there so that we can give a more informative error message. I wasn't the one who wrote this function, though, so not 100% sure. Maybe at some point we should evaluate if the error handling complexity provides adequate benefits in this case?

jdhoffa: Yup maybe just good to make a new issue about it, but it could think about assessing if it's worth it

cjyetman: made a new issue for it #509

PR #523, opened 2023-11-29
Use increasing_or_decreasing instead of green_or_brown

### LINKED ISSUE #520: in plot_trajectory, replace `green_or_brown` with `increasing_or_decreasing`
https://github.com/RMI-PACTA/r2dii.plot/blob/b4b87e8a3fe0a07367b1de15ef097250507f258e/R/plot_trajectory.R#L311C22-L311C48

When using the latest r2dii.data (https://github.com/RMI-PACTA/r2dii.data/pull/309 this or newer), the `plot_trajectory()` function will throw a warning that `green_or_brown` is deprecated. Fix by replacing with `increasing_or_decreasing`

MonikaFu:
Closes #520

### DIFF
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -255,7 +255,7 @@ order_trajectory <- function(data) {
 
   technology_kind <- get_tech_kind(data)
 
-  if (technology_kind == "green") {
+  if (technology_kind == "increasing") {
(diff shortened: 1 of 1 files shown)


cjyetman (approved):
makes sense but didn't test

PR #525, opened 2023-12-08
Remove current scenario bar from techmix

### LINKED ISSUE #513: Remove scenario target in the "current year" of analysis
Making this ticket for George

Screenshot 2023-05-19 at 2 16 21 PM

MonikaFu:
Closes #513 

In this PR I add a filter to techmix prep function so that it filters out the scenario bar for start year. I did not keep the option to display the scenario bar in the current year as it does not make much sense to look at scenario in the start year as it is always equal to the portfolio.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # r2dii.plot (development version)
 
+* techmix plot does not show the start year scenario bar anymore. (#513)
+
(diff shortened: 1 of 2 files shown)


AlexAxthelm (approved):
Makes sense.

cjyetman (approved):

PR #528, opened 2024-01-17
`plot_emission_intensity` correctly orders `scale_colour_r2dii` input

### LINKED ISSUE #527: `plot_emission_intensity` outputs metrics with incorrect order of colours when used with `scale_colour_r2dii`
Expected behaviour:
Using `plot_emission_intensity` with input `data` that has specifically ordered factors for `emission_factor_metric` in conjunction with `scale_colour_r2dii` with custom colour `labels`, outputs a plot whos `emission_factor_metric` has those colours (same order).  

Observed behaviour:
Output has the correct values for colour, however the order gets scrambled.

See reprex below:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(r2dii.plot)

data <- sda %>%
  filter(region == "global")

input_levels <- c(
  "projected",
  "corporate_economy",
  "target_demo",
  "adjusted_scenario_demo"
)

input_color_scale <-
  c(
    "dark_blue",
    "green",
    "grey",
    "ruby_red"
  )

input_color_scale_hex <- data.frame(label = input_color_scale) %>%
  left_join(palette_colours, by = "label") %>%
  pull(hex)

expected_output <- data.frame(
  levels = input_levels,
  hex = input_color_scale_hex
)

p <- data %>%
  dplyr::mutate(
    emission_factor_metric = factor(
      .data$emission_factor_metric,
      levels = input_levels
    )
  ) %>%
  plot_emission_intensity() +
  scale_colour_r2dii(
    labels = input_color_scale
  )
#> Warning: The `data` argument of `plot_emission_intensity()` must be prepped already as
#> of r2dii.plot 0.4.0.
#> ℹ From the next release you will need to call
#>   `r2dii.plot::plot_emission_intensity(data)` prior to calling
#>   `r2dii.plot::plot_emission_intensity()`.
#> ℹ Alternatively custom data preparation will also become possible.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> Scale for colour is already present.
#> Adding another scale for colour, which will replace the existing scale.

# print the levels that colours are applied to
ordered_output_levels <- levels(r2dii.plot:::match_lines_order(p$data))

# print the actual colour scales of the plot
ordered_output_colour_scale <- p$scales$get_scales("colour")$palette(
  length(ordered_output_levels)
)

plot_output <- data.frame(
  levels = ordered_output_levels,
  hex = ordered_output_colour_scale
)

# Note that `corporate_economy` and `projected` are swapped
plot_output
#>                   levels     hex
#> 1      corporate_economy #1b324f
#> 2              projected #00c082
#> 3            target_demo #d0d7e1
#> 4 adjusted_scenario_demo #a63d57
expected_output
#>                   levels     hex
#> 1              projected #1b324f
#> 2      corporate_economy #00c082
#> 3            target_demo #d0d7e1
#> 4 adjusted_scenario_demo #a63d57
```

<sup>Created on 2024-01-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

jdhoffa:
@jacobvjk has pointed out that, if a user passes a custom colour scaling to `plot_emission_intensity`, along with specific levels of `emission_factor_metric` to which the scaling should be applied, the result doesn't appear with correct colours. 

This is caused by an internal function, `match_lines_order` that forcibly changes the levels of the dataset to a custom order (which the user of the function can't know arbitrarily). This means that custom scalings are impossible to achieve consistently, as they will always be applied to the factor levels as ordered within the function. 

This function was introduced to solve (what seems like) a cosmetic problem, and removing it will likely only have the minor consequence of the legend colour orders not being identical the line colour orders, as they appear on the plot (a UI issue, but not a functional one). See discussion in #527 

To fix this:
* I got rid of `match_lines_order`
* I changed an argument name from `labels` to `colour_labels`. The original argument clashed with a required argument name to be passed in through `...`

Closes #527


jdhoffa:
PTAL: We should look into these deprecation warnings, there's something funky going on there, see #530

MonikaFu:
LGTM. Also tested locally and works as expected. As I mentioned in #527 I would like to try to bring back the ordering of the legend (using a different solution that does not impact the order of the data itself). Will you make an issue about it or should I?

MonikaFu (approved):

jdhoffa:
#534 is the issue, feel free to edit it as you see fit :-)

jacobvjk:
@jdhoffa  one comment on this. I understood the test as an implied order of colours mapped to categories. I.e. "projected" should be "dark_blue". Running this locally does not return projected as dark blue.

So just to be sure: does this fix aim to allow mapping the exact pairs of categories to colours outside the function? If so, great. If the idea was to produce the mapping that I would say the test implies, I don't seem to be able to transfer this to the workflow I am using this on.

jdhoffa:
Oh man, ok. Let me have a look tomorrow, maybe i messed something up.

jdhoffa:
@jacobvjk can you try this
``` r
data <- filter(sda, sector == "cement", region == "global") %>% 
  dplyr::mutate(
    emission_factor_metric = factor(
      .data$emission_factor_metric, 
      levels = c(
        "projected", 
        "corporate_economy", 
        "target_demo", 
        "adjusted_scenario_demo"
      )
    )
  )

input_colour_scale <- c(
  "dark_blue",
  "green",
  "grey",
  "ruby_red"
)

p <- plot_emission_intensity(data)

p <- p + scale_colour_r2dii(
  colour_labels = input_color_scale,
)
```

PR #538, opened 2024-01-25
`plot_emission_intensity(..., convert_label = to_title)` plays nicely with `scale_colour_r2dii`

### LINKED ISSUE #536: Calling `plot_emission_intensity(..., convert_lable = to_title)` in conjunction with `scale_colour_r2dii`  yields mis-matched colours
It is a similar problem as described in #535 only here is triggered when using the function argument instead of using the `label` column in data.

jdhoffa:
Closes #536


MonikaFu (approved):
LGTM

PR #541, opened 2024-01-29
Skip tests with new args if R version is older

### LINKED ISSUE #540: Investigate why `ubuntu-latest (oldrel-4)` is failing
https://github.com/RMI-PACTA/r2dii.plot/actions/runs/7693990172/job/20963825156?pr=537

jdhoffa:
This PR skips some tests on checks in older versions of R.

I thought about this, and I think it's actually ok... we are only skipping the testing infrastructure on older versions of R, but the tests still run and must pass before we merge PRs meaning that the functionality still should be as expected. It's only that these older versions of R can't run the tests, but they shouldn't ever have to since those are only dev-facing. 

Closes #540

### DIFF
--- a/R/plot_emission_intensity.R
+++ b/R/plot_emission_intensity.R
@@ -65,8 +65,16 @@ prep_emission_intensity <- function(data,
                                     convert_label = identity,
                                     span_5yr = FALSE) {
   out <- data %>%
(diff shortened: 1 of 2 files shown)


jdhoffa:
I guess I accidentally handled this in a different PR XD

Issue #529, opened 2024-01-23
Why have the design-pattern `function()` with the body `function_impl()`, defined elsewhere?


jdhoffa:
For example, in the file `scale_colour_r2dii.R`, there is a function defined as `r2dii_pal`, that has an abort call and a body of `r2dii_pal_impl` that is defined in a different file `utils.R`.

What are the pros to doing this?

The main cons I notice are that context shifting (and not being able to directly see what a function does when looking at it) takes time and makes it difficult to interpret what a function does.

jdhoffa:
I think the last place this appears is here:
https://github.com/RMI-PACTA/r2dii.plot/blob/ec42cf33fb7d07788ed8513fe753073e9c2e167d/R/utils.R#L282-L291

MonikaFu:
This implementation was done to DRY the code but indeed I agree that it could probably do without both `r2dii_pal()` and `r2dii_pal_impl()` and just one helper function after some refactoring.

PR #543, opened 2024-01-30
Extract `prep_*` functions from `plot_*` functions

### LINKED ISSUE #465: Move prep functions outside of plot functions
Currently, the `plot_*()` functions do more than they promise as they contain data preparation steps as well which are not explicit in the function name. Following KISS software design principle, the data preparation functions will be taken out of the `plot_*()` functions and placed in a separate user-facing function which they can (but don't have to) use for structuring the data. The `qplot_*()` functions remain unchanged.

Desired workflow for using `plot_*()` functions will be:
```r
prepped_data <- prep_trajectory(raw_data, parameters)

plot <- plot_trajectory(prepped_data)
```
or

```r
custom_prepped_data <- user_defined_data_prep(raw_data)

plot <- plot_trajectory(custom_prepped_data)
```

FYI @AlexAxthelm @maurolepore

jdhoffa:
My plan here is to:

- [x] Extract, document and export the `prep_*` functions as-is, without touching the `plot_*` functions. Write tests for them, and ensure everything passes (just a refactor).
- [x] Alter the `plot_*` functions to expect the prepared data (also alter these expectations in the relevant tests)
- [x] Make sure to remove any unused `parameters` in the `plot_*` function
- [x] Complete deprecation process
- [x] Update `NEWS.md`
- [x] Review and check all `FIXME`s 

Closes #465


jdhoffa:
@MonikaFu I am working towards completing this, but I don't understand what the `env` argument for `prep_techmix` achieves. It seems to not be user-facing, but I'm curious how it should be handled. 

https://github.com/RMI-PACTA/r2dii.plot/blob/bc4814146bd0f8668b4d85c40a5b582774141a91/R/plot_techmix.R#L59-L70

MonikaFu:
You're right, it is not supposed to be user-facing. The variable is used in internal functions in order to return more meaningful errors. It passes the environment in which the function was called to the inside of the functions so that the name of variable that was used as for example, `data` argument is known and shown to the user in an error message. It should not be part of `prep_*` functions when they become user-facing.

jdhoffa:
Thanks for the context @MonikaFu , I will have a look at what you did, and come back if I have questions :-)

jdhoffa:
Ooof @MonikaFu this is a huge undertaking it turns out. 
I have done a good chunk of the extraction and refactoring (including updating tests) and I think we are at a functional step. 

I have 2 big TO-DOs left:

- [ ] Complete the deprecation process using the `lifecycle` package
- [ ]  Review and check all FIXMEs (I left two tests empty to test some behaviour that I wasn't sure how to test at the time. Going to come back to those in a moment)

But in the mean-time, when you're feeling better, I would love a quick review already, as a lot of files have been touched.

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # r2dii.plot (development version)
 
 ## Breaking change
+* All `plot_*()` functions have had the data preparation step extracted as 
+`prep_*()` functions (#465).
```
jdhoffa, line 5: Feel free to re-word this. I wonder if it maybe deserves more explanation in the `NEWS.md` as it is a Breaking and user-facing API change.

R/plot_emission_intensity.R
```diff
@@ -1,95 +1,22 @@
...
-#'   * The structure must be like [sda].
-#'   * The column `sector` must have a single value (e.g. "cement").
-#'   * (Optional) If present, the column `label` is used for data labels.
-#' @param span_5yr Logical. Use `TRUE` to restrict the time span to 5 years from
-#'   the start year (the default behavior of `qplot_emission_intensity()`), or use
-#'   `FALSE` to impose no restriction.
-#' @template convert_label
-#' @templateVar fun qplot_emission_intensity
-#' @templateVar value to_title
+#' @param data A data frame like the output of `prep_emission_intensity()`.
```
jdhoffa, line 3: Most `plot_*` functions don't have that many arguments, and most get passed to the `prep_*` functions (you will see a similar trend with the tests.

R/plot_emission_intensity.R
```diff
@@ -1,95 +1,22 @@
...
-#' # `data` must meet documented "Requirements"
-#' data <- subset(sda, sector == "cement" & region == "global")
-#' plot_emission_intensity(data)
-#'
 #' # plot with `qplot_emission_intensity()` parameters
-#' plot_emission_intensity(
-#'   data,
-#'   span_5yr = TRUE,
-#'   convert_label = to_title
-#' )
```
jdhoffa, line 29: Examples had to be hugely refactored. To recreate the `qplot_*` functionality, the arguments have to be passed to the corresponding `prep_*` function. I still thought the info was useful, so I made that the ONLY example. 

I can add more examples if you think it is useful.

R/plot_emission_intensity.R
```diff
@@ -1,95 +1,22 @@
...
-                                    span_5yr = FALSE,
-                                    convert_label = identity) {
-  lifecycle::deprecate_soft(
-    when = "0.4.0",
-    what = "plot_emission_intensity(data = 'must be prepped already')",
-    details = api_warning_details(
-      "prep_emission_intensity",
-      "plot_emission_intensity"
-      )
-  )
```
jdhoffa, line 40: I have removed the deprecation warnings entirely. I have a TODO to re-add an appropriate message (I think? do I even need to?? I need to look into how to do this properly).

R/plot_emission_intensity.R
```diff
@@ -1,95 +1,22 @@
...
-    mutate(
-      year = as.Date(ISOdate(year = .data$year, month = 1L, day = 1L))
-    )
-
-  metrics <- distinct(out, .data$emission_factor_metric)
-  colours <- palette_colours[seq_len(nrow(metrics)), "hex", drop = FALSE]
-  specs <- dplyr::bind_cols(metrics, colours)
-
-  left_join(out, specs, by = metric(data))
+  check_emission_intensity(data, env = env)
```
jdhoffa, line 18: `check_plot_emission_intensity` and `check_prep_emission_intensity` ended up being identical, so I just refactored it out entirely.

I do think it's good to do checks on both functions, since they are both now exported.

R/plot_techmix.R
```diff
@@ -226,22 +151,6 @@ get_technology_colours <- function(data) {
...
-    grepl("(?i)fossil(?-i)[a-zA-Z]+", x)   ~ "fossil fuels",
-    TRUE ~ tolower(x)
-  )
-  # styler: on
-}
-
-extract_scenarios <- function(x) {
-  unique(x[startsWith(x, "target_")])
-}
-
```
jdhoffa, line 244: I was refactoring to ensure all functions that were relevant to `prep_*` or `plot_*` were in the appropriate place.
While doing this, I noticed that these are used by multiple functions so i moved them to `utils.R`

R/plot_trajectory.R
```diff
@@ -268,57 +205,6 @@ order_trajectory <- function(data) {
...
-
-get_tech_kind <- function(data) {
-  technology_kind <- r2dii.data::increasing_or_decreasing %>%
-    filter(.data$technology == unique(data$technology)) %>%
-    pull(.data$increasing_or_decreasing) %>%
-    unique()
-
-  technology_kind
-}
-
```
jdhoffa, line 321: All of these are now relevant to `prep_trajectory`

R/plot_trajectory.R
```diff
@@ -352,67 +238,10 @@ get_ordered_scenario_colours <- function(n) {
...
-  data_worse_than_scenarios <- tibble(
-    year = unique(data$year),
-    technology = unique(data$technology),
-    sector = unique(data$sector)
+         "2" = pick(c("light_green", "red")),
+         "3" = pick(c("light_green", "light_yellow", "red")),
+         "4" = pick(c("light_green", "dark_yellow", "light_yellow", "red")),
+         "5" = scenario_colours,
+         abort(c("`n` must be between 2 and 5.", x = glue("Provided: {n}."))) # nocov
   )
```
jdhoffa, line 246: git diff is getting confused here, this was a pure refactor

R/plot_trajectory.R
```diff
@@ -352,67 +238,10 @@ get_ordered_scenario_colours <- function(n) {
...
-        metric = factor(
-          .data$metric,
-          levels = rev(get_ordered_scenarios(data_scenarios))
-        )
-      ) %>%
-      arrange(.data$year, .data$metric) %>%
-      mutate(value_low = lag(.data$value, n = 1, default = area_borders$lower))
-  }
-
-  data_scenarios
```
jdhoffa, line 417: This logic was left alone, and purely refactored to `prep_trajectory`

R/qplot_emission_intensity.R
```diff
@@ -18,7 +18,7 @@
 #' qplot_emission_intensity(data)
 qplot_emission_intensity <- function(data) {
   env <- list(data = substitute(data))
-  check_plot_emission_intensity(data, env = env)
+  check_emission_intensity(data, env = env)
```
jdhoffa, line 21: See above: pure refactor

R/qplot_techmix.R
```diff
@@ -30,8 +30,7 @@ qplot_techmix <- function(data) {
     prep_techmix(
       convert_label = recode_metric_techmix,
       span_5yr = TRUE,
-      convert_tech_label = spell_out_technology,
-      env = env
```
jdhoffa, line 34: `env` is no longer a user-facing argument of `prep_techmix`

R/qplot_trajectory.R
```diff
@@ -63,3 +63,17 @@ labs_trajectory <- function(p, data) {
...
+  hint_if_missing_names(abort_if_missing_names(data, crucial), "market_share")
+  abort_if_has_zero_rows(data, env = env)
+  enforce_single_value <- c("sector", "technology", "region", "scenario_source")
+  abort_if_multiple(data, enforce_single_value, env = env)
+  abort_if_invalid_scenarios_number(data)
+  abort_if_too_many_lines(max = 4, summarise_max_year_by_scenario(data))
+  abort_if_too_many_lines(max = 5, summarise_max_year_by_traj_metric(data))
+
+  invisible(data)
+}
```
jdhoffa, line 79: Added a new check, since `qplot_trajectory` now needs the check conditions for both `prep` and `plot`

tests/testthat/test-plot_emission_intensity.R
```diff
@@ -31,44 +31,18 @@ test_that("outputs an object with no factor-columns derived from `specs`", {
...
-  data <- filter(sda, sector == "cement")
-  abort_if_year_range_is_5yr_already(data)
-
-  p_f <- plot_emission_intensity(data, span_5yr = FALSE)
-  expect_false(diff(year_range(p_f)) == 5)
-
-  p_t <- plot_emission_intensity(data, span_5yr = TRUE)
-  expect_true(diff(year_range(p_t)) == 5)
-})
-
```
jdhoffa, line 69: Many tests were moved to the more appropriate place (e.g. if "too many line" handling is in the `prep_*` function, it has been moved there.

tests/testthat/test-plot_emission_intensity.R
```diff
@@ -80,14 +54,6 @@ test_that("with n metrics in input outputs n lines", {
...
 options(warn = 0)
 
-test_that("throws expected warning about API change", {
-  data <- head(filter(sda, sector == "cement"))
-  expect_snapshot_error(
-    plot_emission_intensity(data),
-    class = "warning"
-  )
-})
-
```
jdhoffa, line 90: All "API change" warning tests have been removed. See my "figure out deprecation" TODO

tests/testthat/test-plot_techmix.R
```diff
@@ -4,49 +4,9 @@ test_that("without a `data` frame errors gracefully", {
...
-  bad_scenario_source$scenario_source <- c("a", "b")
-  expect_snapshot_error(plot_techmix(bad_scenario_source))
-})
-
-test_that("with too few scenarios errors gracefully", {
-  too_few <- head(market_share, 2L)
-  too_few$metric <- c(main_line(), "corporate_economy")
-  expect_snapshot_error(plot_techmix(too_few))
-})
-
```
jdhoffa, line 47: Tests migrated to `prep`

jdhoffa:
Regarding the deprecation process, my assessment is the following:
* We have adequately warned users that we plan to change the API of an existing function (a breaking change)
* Since we are not yet in version `1.0.0` of `r2dii.plot`, we have still allowed ourselves the opportunity to have breaking changes
* We will need to do a minor version bump for sure, as per SemVer
* I do not think we need any further deprecation messages (I think it's fine to simply remove them all)

tests/testthat/test-plot_trajectory.R
```diff
@@ -10,48 +10,51 @@ test_that("isn't restricted to plotting only 5 years", {
...
-  bad_region <- head(market_share, 2L)
-  bad_region$region <- c("a", "b")
-  expect_error(plot_trajectory(bad_region), "bad_region")
+  #FIXME: Figure out this test
+  # # Keep even if already tested in qplot_. Non-standard evaluation is fragile
+  # bad_region <- head(market_share, 2L) %>%
+  #   mutate(region = c("a", "b")) %>%
+  #   prep_trajectory()
+  #
+  # expect_error(plot_trajectory(bad_region), "bad_region")
```
jdhoffa, line 51: I will need some support here. I seem to have broken this test (it no longer errors as expected), and I am sure it has to do with the `.env` argument. Would love to pair with you for ~20min to try to figure this bit out :-)

jdhoffa:
There is a NOTE when running R CMD check:
"Namespace in Imports field not imported from: ‘forcats’ All declared Imports should be used."

I think this note isn't related to this PR, and should be tackled in a separate issue/ PR.

R/plot_emission_intensity.R
MonikaFu: I'd say it would be great if we could add an informative error message in case the user calls plot on unprepped data.  Also - maybe we should write a blog post explaining how the functions work now and why we introduced this change? I guess I could write it if needed.

jdhoffa: We don't have a blog currently :-D lol

tests/testthat/_snaps/plot_techmix.md
```diff
@@ -1,61 +1,9 @@
...
-
-    `bad_sector` must have a single value of `sector`.
-    i Do you need to pick one value? E.g. pick 'a' with: `subset(bad_sector, sector == 'a')`.
-    x Provided: a, b.
-
-# with too many regions errors gracefully
-
-    `bad_region` must have a single value of `region`.
-    i Do you need to pick one value? E.g. pick 'a' with: `subset(bad_region, region == 'a')`.
-    x Provided: a, b.
```
MonikaFu, line 31: These are still relevant I'd say. Or do we move it all to `prep_()` and add a check for if the data have been prepped instead?

tests/testthat/_snaps/plot_techmix.md
```diff
@@ -1,61 +1,9 @@
...
-
-    `bad_scenario_source` must have a single value of `scenario_source`.
-    i Do you need to pick one value? E.g. pick 'a' with: `subset(bad_scenario_source, scenario_source == 'a')`.
-    x Provided: a, b.
-
-# with too few scenarios errors gracefully
-
-    `too_few$metric` must have one scenario.
-    x It has none.
-
```
MonikaFu, line 43: Still relevant I'd say?

tests/testthat/_snaps/prep_emission_intensity.md
```diff
@@ -0,0 +1,6 @@
+# with too many lines to plot errors gracefully
```
MonikaFu, line 1: I'd say this is more of a plot check than a prep check still.

tests/testthat/test-plot_emission_intensity.R
MonikaFu: We need to update the message or move the check I'd say. It is not very intuitive to have a message about too many lines to plot in a function that is not plot.

tests/testthat/test-plot_techmix.R
```diff
@@ -4,49 +4,9 @@ test_that("without a `data` frame errors gracefully", {
   expect_error(plot_techmix(1), "data.frame.*not")
 })
 
-test_that("without `market_share` data errors gracefully", {
-  bad_kind <- filter(sda, sector == first(sector))
-  expect_snapshot_error(plot_techmix(bad_kind))
-})
-
-test_that("with zero-row data errors gracefully", {
```
MonikaFu, line 12: I guess we should keep this test here as well? Someone might pass a zero row data to plot still.

tests/testthat/test-plot_techmix.R
MonikaFu: potentially we might want to keep them in both? Not sure.

tests/testthat/test-plot_trajectory.R
MonikaFu: let's

MonikaFu (changes requested):
Thanks for your work! See my comments. Happy to discuss verbally as well.

MonikaFu:
@jdhoffa see my review on the draft. Also - given these changes we need to update the documentation on the package website. I'd say let's do it in a separate PR because this one is already huge. Also happy to do it myself. Should we open an issue for it?

R/plot_emission_intensity.R
```diff
@@ -1,95 +1,22 @@
...
+  plot_emission_intensity_impl(data)
```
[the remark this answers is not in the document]
jdhoffa: Un-related: This would also close #529 :-)

R/plot_trajectory.R
```diff
@@ -26,57 +14,32 @@
...
-#'   span_5yr = TRUE,
-#'   convert_label = recode_metric_trajectory,
-#'   center_y = TRUE,
 #'   value_col = "percentage_of_initial_production_by_scope",
 #'   perc_y_scale = TRUE
 #' )
 plot_trajectory <- function(data,
-                            span_5yr = FALSE,
-                            convert_label = identity,
-                            center_y = FALSE,
```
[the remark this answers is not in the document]
jdhoffa: Got it!

R/plot_emission_intensity.R
jdhoffa: I agree in principle that the `checks` likely should be different as they test different input. However in practice, the were checking the presence of the exact same columns. 

jdhoffa: Would you like me to split them back into `check_plot_emission_intensity` and `check_prep_emission_intensity` here? Recognizing that they would just be wrapping the same checks?

R/plot_trajectory.R
jdhoffa: I believe this is done

R/plot_trajectory.R
jdhoffa: I believe this is done

R/plot_emission_intensity.R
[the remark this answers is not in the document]
MonikaFu: yup! Two birds with one stone or something

R/plot_emission_intensity.R
MonikaFu: They used to check for the same things because they were both using unprepped data. But now since `plot` requires prep data I'd expect that it's checks should change and check if the data passed to it indeed is prepped.

jdhoffa: Will be handled in #548

NEWS.md
jdhoffa: @MonikaFu to reword this if you like!

R/plot_emission_intensity.R
jdhoffa: I wonder if this should also be handled in #548 ?

R/qplot_trajectory.R
jdhoffa: NIT: I actually think that may be a bit of a dangerous assumption. I tend to think that any exported function should have it's own input checks. 

Consider the (unlikely) case that we decide to deprecate the `prep_*` and `plot_*` functions, and decide to remove all of the checks from them. `qplot_*` would not realize this, and would be left checkless.

But of course, you are maintainer, and if you would prefer I remove the checks here I am happy to do so!

tests/testthat/test-plot_emission_intensity.R
jdhoffa: Ok so, to move this test to `plot_emission_intensity`, I had to update the checks (so that the lin check only occurs on plot, which totally makes sense). 

Doing so split the `check_*_emission_intensity` function up again, as there is now 1 additional check that occurs only for `plot_emission_intensity`. However, I will leave #548 open, as I imagine we should still add many checks  to prove that `plot_*` isn't accepting un-prepared data...

tests/testthat/test-plot_techmix.R
jdhoffa: I think this should be handled in #548, when we add more checks to the`prep_*` and `plot_*` functions (we can also add tests that ensure each check functions as expected.

jdhoffa:
Probably closes #529

tests/testthat/test-plot_techmix.R
jdhoffa: I think we should handle this also in #548 when we add more checks (and corresponding tests)

jdhoffa:
There is a R CMD CHECK note that I believe is unrelated to this PR. Relates to #547

R/qplot_trajectory.R
jdhoffa: Closing a resolved, see all the `qplot_` functions now :-)

R/plot_emission_intensity.R
MonikaFu: Yes, I agree, it should be part of the checks in #548 .

R/qplot_emission_intensity.R
```diff
@@ -18,11 +18,18 @@
 #' qplot_emission_intensity(data)
 qplot_emission_intensity <- function(data) {
   env <- list(data = substitute(data))
+  check_prep_emission_intensity(data, env = env)
```
MonikaFu, line 21: just to confirm - you are calling the checks here (there are also inside `prep_emission_intensity` in order to be able to pass the right `env` to it? Maybe long term it should be handled in a more elegant way. Now we are calling the checks twice.

R/plot_trajectory.R
```diff
@@ -26,94 +18,44 @@
...
 #' plot_trajectory(
 #'   data,
-#'   span_5yr = TRUE,
-#'   convert_label = recode_metric_trajectory,
 #'   center_y = TRUE,
-#'   value_col = "percentage_of_initial_production_by_scope",
 #'   perc_y_scale = TRUE
 #' )
 plot_trajectory <- function(data,
                             span_5yr = FALSE,
```
MonikaFu, line 37: `span_5yr` should be part of `prep_trajectory`. The same as for `emission_intensity` and `techmix`.

MonikaFu (changes requested):

MonikaFu:
@jdhoffa I see checks failing also due to changes in `r2dii.analysis`. It seems like there was a name change of some of output columns. We should not solve it in this PR but it should be solved as the demo datasets in `r2dii.plot` should follow the output of `r2dii.analysis`.

R/qplot_emission_intensity.R
jdhoffa: Correct! 
NIT: For safety, I do believe calling input checks explicitly on any exported function is prudent (if inelegant), for reasons stated in a different comment. But leave it to you to manage as you see fit :-)

jdhoffa:
> @jdhoffa I see checks failing also due to changes in `r2dii.analysis`. It seems like there was a name change of some of output columns. We should not solve it in this PR but it should be solved as the demo datasets in `r2dii.plot` should follow the output of `r2dii.analysis`.

That is unexpected to me! I will look into it in `r2dii.analysis` and get back to you.

jdhoffa:
> @jdhoffa I see checks failing also due to changes in `r2dii.analysis`. It seems like there was a name change of some of output columns. We should not solve it in this PR but it should be solved as the demo datasets in `r2dii.plot` should follow the output of `r2dii.analysis`.

tl;dr: Broken checks are unrelated to this PR.

In case you are interested:

The failed checks are coming from a "breaking change" in `r2dii.data`, where we have changed the expected format of the NACE codes (cc @jacobvjk)

* The breaking change happened here (where we changed how we format NACE codes), this introduced the bug: https://github.com/RMI-PACTA/r2dii.data/pull/327
* This PR should fix it, by aligning `loanbook_demo` with the new `NACE` code: https://github.com/RMI-PACTA/r2dii.data/pull/338

R/plot_trajectory.R
jdhoffa: Done!

jdhoffa:
I will wait until https://github.com/RMI-PACTA/r2dii.data/pull/338 is merged before re-requesting a review, so that we make sure CI/CD works well.

R/plot_trajectory.R
jdhoffa: Done

jdhoffa:
All checks now passing, this is ready for review!

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # r2dii.plot (development version)
 
 ## Breaking change
+* All `plot_*()` functions have had the data preparation step extracted into 
+`prep_*()` functions (#465).
```
MonikaFu, line 5:
```suggestion
* All `plot_*()` functions have had the data preparation step extracted into 
`prep_*()` functions. This means that from now on `prep_*()` function needs
to be called on `data` prior to `plot_*()`. The APIs of `qplot_*()` functions stay 
unchanged (#465).
```

MonikaFu (approved):
LGTM. See my proposed change to `NEWS.md`.

MonikaFu (approved):
LGTM

jdhoffa:
Going to wait for confirmation that this won't affect ongoing projects, then I will merge :-)

jdhoffa:
Confirmed by @daisy-pacheco and @jacobvjk
Merging!

PR #551, opened 2024-02-20
Add more specific input checks for `plot_emission_intensity` and `prep_emission_intensity`

### LINKED ISSUE #546: Split `check_emission_intensity` into `check_prep_emission_intensity` and `check_plot_emission_intensity`
Also requires adding additional checks to both functions, since it is expected that their inputs should differ dramatically (so likely the checks should too)

              I agree in principle that the `checks` likely should be different as they test different input. However in practice, the were checking the presence of the exact same columns. 

_Originally posted by @jdhoffa in https://github.com/RMI-PACTA/r2dii.plot/pull/543#discussion_r1488101565_

jdhoffa:
The crucial difference between the input to `prep_emission_intensity` and the input to `plot_emission_intensity` (that I can see) is that the latter has the additional `label` column.

There is an "optional" difference, dependent on the value of `span_5yr`, but I don't think it makes sense to check for that since `plot_*` can't know a-priori what the argument passed to `prep_*` was.

The rest of the checks (that the input data is a data-frame etc.) are shared between the two functions.

Note: I wonder, perhaps we should add an exclusion test that the column `label` is NOT in the input to `prep_emission_intenstiy()`. @MonikaFu let me know if you have thoughts there.

Note: Adding the check brought a few bugs in the unit tests to light! woo 👯

Maybe closes #546 
Relates to #548

### DIFF
--- a/R/plot_emission_intensity.R
+++ b/R/plot_emission_intensity.R
@@ -39,10 +39,10 @@ plot_emission_intensity <- function(data) {
 }
 
 check_plot_emission_intensity <- function(data, env) {
+  check_prep_emission_intensity(data, env)
   crucial <- c(prep_emission_factor_crucial, "label")
   hint_if_missing_names(abort_if_missing_names(data, crucial), "sda")
-  check_prep_emission_intensity(data, env)
-  stopifnot(is.data.frame(data))
   abort_if_too_many_lines(data, max = 7)
+
   invisible(data)
 }
--- a/R/plot_emission_intensity.R
+++ b/R/plot_emission_intensity.R
@@ -39,6 +39,8 @@ plot_emission_intensity <- function(data) {
 }
 
 check_plot_emission_intensity <- function(data, env) {
+  crucial <- c(prep_emission_factor_crucial, "label")
+  hint_if_missing_names(abort_if_missing_names(data, crucial), "sda")
   check_prep_emission_intensity(data, env)
   stopifnot(is.data.frame(data))
   abort_if_too_many_lines(data, max = 7)
--- a/tests/testthat/_snaps/plot_emission_intensity.md
+++ b/tests/testthat/_snaps/plot_emission_intensity.md
@@ -11,15 +11,6 @@
     ! `data` must have all the expected names.
     x Missing names: emission_factor_metric, emission_factor_value, label.
 
-# if `data` has zero rows errors gracefully
-
-    `data` must have all the expected names.
-    x Missing names: label.
-    i Is your data `sda`-like?
-    Caused by error in `abort_if_missing_names()`:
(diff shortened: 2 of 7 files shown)


R/plot_emission_intensity.R
```diff
@@ -39,6 +39,8 @@ plot_emission_intensity <- function(data) {
 }
 
 check_plot_emission_intensity <- function(data, env) {
+  crucial <- c(prep_emission_factor_crucial, "label")
+  hint_if_missing_names(abort_if_missing_names(data, crucial), "sda")
   check_prep_emission_intensity(data, env)
```
jdhoffa, line 44: I am leveraging here all of the checks for `prep_emission_intensity`.
Perhaps that function should also gain additional checks?

tests/testthat/_snaps/plot_emission_intensity.md
```diff
@@ -1,26 +1,24 @@
...
+    x Missing names: emission_factor_metric, emission_factor_value, label.
     i Is your data `sda`-like?
     Caused by error in `abort_if_missing_names()`:
     ! `data` must have all the expected names.
-    x Missing names: emission_factor_metric, emission_factor_value.
-
-# if `data` has zero rows errors gracefully
-
-    `zero_row` must have some rows.
-    x `zero_row` has zero rows.
```
jdhoffa, line 17: This test moved to `prep_*`

Perhaps a similar test should happen in `plot_*`?

jdhoffa: Done: I added the same test to both functions.

tests/testthat/test-plot_emission_intensity.R
```diff
@@ -10,7 +10,7 @@ test_that("if `data` is not sda-like errors gracefully", {
 })
 
 test_that("if `data` has zero rows errors gracefully", {
-  zero_row <- sda[0L, ]
+  zero_row <- prep_emission_intensity(sda)[0L, ]
```
jdhoffa, line 13: This was a bug that the check brought to light! 👯

MonikaFu:
Thanks @jdhoffa. Reviewing now. 

As for this

>Note: I wonder, perhaps we should add an exclusion test that the column label is NOT in the input to prep_emission_intenstiy(). @MonikaFu let me know if you have thoughts there.

This check is not necessary. The `label` column is an optional column for the input. Function works with and without it.

MonikaFu:
>There is an "optional" difference, dependent on the value of span_5yr, but I don't think it makes sense to check for that since plot_* can't know a-priori what the argument passed to prep_* was.

I agree. We don't need to test for that.

MonikaFu (approved):
LGTM. Also tested locally.

PR #552, opened 2024-02-21
`plot_techmix` and `plot_trajectory` gain more specific checks

### LINKED ISSUE #548: Ensure all `prep_*` and `plot_*` functions have individualized (and different) checks
I agree in principle that the `checks` likely should be different as they test different input. However in practice, the were checking the presence of the exact same columns. 

_Originally posted by @jdhoffa in https://github.com/RMI-PACTA/r2dii.plot/pull/543#discussion_r1488101565_

jdhoffa:
* `plot_*` functions gain check for the presence of the `label` column
* Brought to light a bug in a `qplot` function that was fixed
* stray `env` arg removed from a `prep_` function that didn't need it

Closes #548

### DIFF
--- a/R/prep_trajectory.R
+++ b/R/prep_trajectory.R
@@ -33,9 +33,7 @@ prep_trajectory <- function(data,
                             span_5yr = FALSE,
                             value_col = "percentage_of_initial_production_by_scope") {
 
-  env <- list(data = substitute(data))
-
-  check_prep_trajectory(data, value_col = value_col, env = env)
+  check_prep_trajectory(data, value_col = value_col)
 
   data <- data %>%
     prep_common() %>%
@@ -49,7 +47,7 @@ prep_trajectory <- function(data,
   data
 }
 
-check_prep_trajectory <- function(data, value_col, env) {
+check_prep_trajectory <- function(data, value_col) {
   stopifnot(is.data.frame(data))
   crucial <- c(common_crucial_market_share_columns(), value_col)
   hint_if_missing_names(abort_if_missing_names(data, crucial), "market_share")
--- a/R/plot_techmix.R
+++ b/R/plot_techmix.R
@@ -75,7 +75,12 @@ plot_techmix <- function(data) {
 
 check_plot_techmix <- function(data, env) {
   stopifnot(is.data.frame(data))
-  crucial <- c(common_crucial_market_share_columns(), "technology_share")
+  crucial <- c(
+    common_crucial_market_share_columns(),
+    "technology_share",
+    "label",
+    "label_tech"
+    )
   hint_if_missing_names(abort_if_missing_names(data, crucial), "market_share")
   abort_if_has_zero_rows(data, env = env)
   enforce_single_value <- c("sector", "region", "scenario_source")
--- a/R/plot_trajectory.R
+++ b/R/plot_trajectory.R
@@ -219,6 +219,8 @@ start_value_portfolio <- function(data) {
 
 check_plot_trajectory <- function(data, env) {
   stopifnot(is.data.frame(data))
+  crucial <- c(common_crucial_market_share_columns(), "label")
+  hint_if_missing_names(abort_if_missing_names(data, crucial), "market_share")
   abort_if_has_zero_rows(data, env = env)
   enforce_single_value <- c("sector", "technology", "region", "scenario_source")
   abort_if_multiple(data, enforce_single_value, env = env)
--- a/R/qplot_techmix.R
+++ b/R/qplot_techmix.R
@@ -24,7 +24,7 @@
 #' qplot_techmix(data)
 qplot_techmix <- function(data) {
   env <- list(data = substitute(data))
-  check_plot_techmix(data, env = env)
+  check_prep_techmix(data, env = env)
 
   data %>%
     prep_techmix(


R/prep_trajectory.R
```diff
@@ -33,9 +33,7 @@ prep_trajectory <- function(data,
                             span_5yr = FALSE,
                             value_col = "percentage_of_initial_production_by_scope") {
 
-  env <- list(data = substitute(data))
-
-  check_prep_trajectory(data, value_col = value_col, env = env)
```
MonikaFu, line 38: We're missing a couple of checks on data requirements (see lines 5-6 in this file). Once those are implemented the `env` argument will be needed for `check_prep_trajectory()`.

In principle the minimum when it comes to checks is checking if all requirements for parameters described in function documentation are met.

R/prep_trajectory.R
```diff
@@ -49,7 +47,7 @@ prep_trajectory <- function(data,
   data
 }
 
-check_prep_trajectory <- function(data, value_col, env) {
```
MonikaFu, line 52: see comment above

MonikaFu (changes requested):
See my comments regarding `prep_trajcetory()`. It might also be a good idea to go through the documentation of each function `prep_*()`, `plot_*()`, `qplot_*` for both `trajectory` and `techmix` and see if we have checks for each requirement listed there.

jdhoffa:
Note to self: go through the documentation of all functions to ensure sufficient input checks:

- [x] `prep_trajectory` and `plot_trajectory`
- [ ] `prep_techmix` and `plot_techmix`
- [ ] `prep_emission_intensity` and `plot_emission_intensity`

jdhoffa:
@MonikaFu All documented expectations should now be well represented in code for all `prep`, `plot` and `qplot` style functions.

except for: 
https://github.com/RMI-PACTA/r2dii.plot/blob/5210ebcd5b44e5dabee12e6b00eff121f054c80f/R/prep_techmix.R#L7-L8

There is a check for single scenarios, but I haven't seen any checks anywhere for the presence of "a portfolio and benchmark" on input. Does that part of the documentation require a "bare minimum" check or is it an exception?

Note: The package also had no checks for the input parameters `convert_label` or `span_5yr` on any of the functions, so I added some very simple checks there. You may improve them in a future PR if you like, but at least they are covered now.

Note: I was waffling back and forth between how to refactor the input checks of the `qplot_` functions (since they really should be identical to the requirements to the `prep_` functions, just with default arguments). If you have a preference there, let me know.

MonikaFu:
> There is a check for single scenarios, but I haven't seen any checks anywhere for the presence of "a portfolio and benchmark" on input. Does that part of the documentation require a "bare minimum" check or is it an exception?

Indeed, it should be checking for presence of "projected" value in `metric` column. We missed that. The benchmark is optional. I propose to reword it in the documentation to:

* The column `metric` must have a "projected" value and 
a single `scenario` (e.g. "target_sds").
* (Optional) The column `metric` may contain a benchmark 
 (e.g. "corporate_economy").

And check if "projected" is in the metric column. The presence of scenario is checked in `abort_if_multiple_scenarios()`. To be consistent with what the function actually does we should maybe rename it to `abort_if_wrong_number_of_scenarios()` or something like that.

> Note: The package also had no checks for the input parameters convert_label or span_5yr on any of the functions, so I added some very simple checks there. You may improve them in a future PR if you like, but at least they are covered now.

Thanks.

> Note: I was waffling back and forth between how to refactor the input checks of the qplot_ functions (since they really should be identical to the requirements to the prep_ functions, just with default arguments). If you have a preference there, let me know.

Your implementation makes sense to me.

> Note: Windows checks seem to be failing due to the breaking change to r2dii.data that introduced the new increasing_or_decreasing dataset.

This is strange because `r2dii.plot` is already using `increasing_or_decreasing` from `r2dii.data`. Unless something else changed.

jdhoffa:
> Indeed, it should be checking for presence of "projected" value in `metric` column. We missed that. The benchmark is optional. I propose to reword it in the documentation to:
> 
> * The column `metric` must have a "projected" value and
>   a single `scenario` (e.g. "target_sds").
> * (Optional) The column `metric` may contain a benchmark
>   (e.g. "corporate_economy").

Awesome. Given that this PR is meant to only update checks (a `maintenance` update), and changing the docs would be user-facing, I would prefer to handle it in a separate PR. See new issue #557 

> And check if "projected" is in the metric column. The presence of scenario is checked in `abort_if_multiple_scenarios()`. To be consistent with what the function actually does we should maybe rename it to `abort_if_wrong_number_of_scenarios()` or something like that.

- [ ] I will look into this. Would you like me to also re-name the check? Is `abort_if_wrong_number_of_scenarios()` the name you prefer?

> This is strange because `r2dii.plot` is already using `increasing_or_decreasing` from `r2dii.data`. Unless something else changed.

Indeed. I don't believe it is related to this PR, but I will look into it later. See #556

MonikaFu:
> Awesome. Given that this PR is meant to only update checks (a maintenance update), and changing the docs would be user-facing, I would prefer to handle it in a separate PR. See new issue https://github.com/RMI-PACTA/r2dii.plot/issues/557
Sure

`abort_if_wrong_number_of_scenarios()` sounds good to me unless you have an idea for sth shorter.

jdhoffa:
@MonikaFu this is ready for re-review

MonikaFu (approved):
LGTM

PR #579, opened 2025-02-06
modernize pkgdown site in line with other r2dii packages

cjyetman:
Similar to what was done in https://github.com/RMI-PACTA/r2dii.match/pull/499, this modernizes the pkgdown config and fixes some problems, e.g. the text being black instead of white in the current dev version of the site at https://rmi-pacta.github.io/r2dii.plot/dev/

note: authors section is removed from pkgdown config because authors are automatically inferred from DESCRIPTION

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -14,36 +14,29 @@ home:
   - text: Learn more
     href: https://rmi.org/
 navbar:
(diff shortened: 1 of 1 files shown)


MonikaFu (approved):
LGTM

PR #580, opened 2025-02-06
add known non-spelling mistakes to `inst/WORDLIST`

cjyetman:


inst/WORDLIST
```diff
@@ -1,22 +1,24 @@
...
 colours
 dii
 funder
+funders
 gg
 ggplot
-https
 pts
 rect
-sda
```
MonikaFu, line 19: why are we removing the two above? Is it not safer to keep them in?

MonikaFu (approved):
NIT comment but overall LGTM.

inst/WORDLIST
cjyetman: It's an automated process, so I guess they use to exist but don't anymore.

cjyetman:
@MonikaFu had to bump the dev version number again because there were multiple PRs queued up

jdhoffa (dismissed):

jdhoffa (approved):

PR #582, opened 2025-02-06
Update readme images

cjyetman:
not sure when/how these got out of sync, but every time I run `devtools::build_readme()` it wants to modify these images


MonikaFu (approved):
LGTM

PR #575, opened 2025-01-31
Compatibility with ggplot2 3.6.0

teunbrand:
Hi there,

Apologies for not posting an issue first.
The ggplot2 package is planning an update for around May 2025 and a reverse dependency test identified a problem with the RMI-PACTA package.
The details are explained in https://github.com/tidyverse/ggplot2/issues/6290, but essentially ggplot2 doesn't populate the `plot$labels` field before plot building anymore, which violates some test assumptions in this package.

This PR changes the tests to be compatible with both versions of ggplot2. 
You can test the changes yourself with the development version of ggplot2 (`pak::pak("tidyverse/ggplot2")`)

Best,
Teun

PR #588, opened 2025-02-10
don't use `scale_name` with {ggplot2} v3.5.0+

### LINKED ISSUE #577: Maintenance: The `scale_name` argument of `discrete_scale()` is deprecated as of ggplot2 3.5.0.
When running the tests with the dev version of `ggplot2` a deprecation warning appeared during testing. Update the code to prevent the warning.

cjyetman:
- resolves #577 

`scale_name` is [deprecated as of {ggplot2} v3.5.0](https://ggplot2.tidyverse.org/news/index.html#breaking-changes-3-5-0)

used the same strategy as [here](https://github.com/nanxstats/ggsci/pull/26) and [here](https://github.com/nanxstats/ggsci/pull/30) to adapt but remain backward compatible with {ggplot2} < v3.5.0

⚠️ note: @MonikaFu I didn't see the original deprecation warning mentioned in #577, but I now see messages in the tests about a colour scale being replaced 🤷🏻 
> Scale for colour is already present.
> Adding another scale for colour, which will replace the existing scale.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: r2dii.plot
 Title: Visualize the Climate Scenario Alignment of a Financial Portfolio
-Version: 0.4.0.9009
+Version: 0.4.0.9010
 Authors@R: 
     c(person(given = "Monika",
              family = "Furdyna",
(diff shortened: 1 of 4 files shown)


MonikaFu:
@cjyetman indeed, if you have CRAN version of the package you will not see the deprecation message. I saw it when I installed `dev` version of `ggplot2` while reviewing the external PR #575

PR #595, opened 2025-02-13
set values for doc template in `prep_trajectory()`

### LINKED ISSUE #594: potential bug: `devtools::document()` errors with hard-to-interpret message
Both @MonikaFu  and I get an error when running devtools::document(), althought the messages vary slightly. In neither case are the messages particularly insightful.

My reprex:

```r
devtools::load_all(path = "/path/to/git/r2dii.plot")
#> ℹ Loading r2dii.plot
devtools::document(pkg = "/path/to/git/r2dii.plot")
#> ℹ Updating r2dii.plot documentation
#> ℹ Loading r2dii.plot
#> Error in eval(code, envir = envir) : object 'fun' not found

```
<sup>Created on 2025-02-13 with [reprex v2.1.0](https://reprex.tidyverse.org/) </sup>

Note: I changed my local path to `"/path/to/git/r2dii.plot"` in this issue

cjyetman:
- resolves #594
- also fixes this section in the documentation https://rmi-pacta.github.io/r2dii.plot/reference/prep_trajectory.html#arg-convert-label

using
https://github.com/RMI-PACTA/r2dii.plot/blob/31331f32690c7765b140c0554eb8446945c8f26c/R/prep_trajectory.R#L8

requires setting `fun` and `value` because of the parameters used here
https://github.com/RMI-PACTA/r2dii.plot/blob/31331f32690c7765b140c0554eb8446945c8f26c/man-roxygen/convert_label.R#L1-L5

⚠️ note that using `@template` and `@templateVars` appears to be [superseded and no longer recommended](https://roxygen2.r-lib.org/articles/reuse.html#superseded)

### DIFF
--- a/R/prep_trajectory.R
+++ b/R/prep_trajectory.R
@@ -6,6 +6,8 @@
 #' `region`, `scenario_source`.
 #' * (Optional) If present, the column `label` is used for data labels.
 #' @template convert_label
+#' @templateVar fun qplot_trajectory
+#' @templateVar value recode_metric_trajectory
 #' @param span_5yr Logical. Use `TRUE` to restrict the time span to 5 years from
 #'   the start year (the default behavior of `qplot_trajectory()`), or use
 #'   `FALSE` to impose no restriction.
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: r2dii.plot
 Title: Visualize the Climate Scenario Alignment of a Financial Portfolio
-Version: 0.5.0.9000
+Version: 0.5.0.9001
 Authors@R: 
     c(person(given = "Monika",
              family = "Furdyna",
--- a/man/prep_trajectory.Rd
+++ b/man/prep_trajectory.Rd
@@ -24,7 +24,8 @@ prep_trajectory(
 y-axis labels. For example:
 \itemize{
 \item To convert labels to uppercase use \code{convert_label = toupper}.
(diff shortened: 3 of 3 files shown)


R/prep_trajectory.R
```diff
@@ -6,6 +6,8 @@
 #' `region`, `scenario_source`.
 #' * (Optional) If present, the column `label` is used for data labels.
 #' @template convert_label
+#' @templateVar fun qplot_trajectory
+#' @templateVar value recode_metric_trajectory
```
cjyetman, line 10: note: I got this from here (which I assume is the correct thing to do)
https://github.com/RMI-PACTA/r2dii.plot/blob/31331f32690c7765b140c0554eb8446945c8f26c/R/qplot_trajectory.R#L35-L36

jacobvjk (approved):
lgtm, as far as I can judge it. `devtools` plays nice again, not sure about the changed content. That would be for @MonikaFu to validate

R/prep_trajectory.R
cjyetman: @MonikaFu can you verify this please?

MonikaFu: Yes, this is correct!

MonikaFu (approved):
LGTM. Also, thanks for the flag @cjyetman that the solution we're using is no longer recommended by roxygen. Not sure if we'll be able to look into it in the near future but maybe worth creating an issue about it <3

PR #605, opened 2025-03-03
use subscript 2 in CO₂ in emission intensity plot axis label

cjyetman:


MonikaFu (approved):
LGTM

cjyetman:
@jdhoffa do you see any problem with this?

jdhoffa:
Nope! This is legit.

Anything that is just changing how the value ends up getting presented (documentation, plots, etc.) is fine. 
Just changing the underlying data would probably cause issues

PR #607, opened 2025-03-15
remove re-exported `%>%` and adjust examples

### LINKED ISSUE #583: remove re-exported pipe
I can't really figure out why `%>%` was ever re-exported from `magrittr` here, but it doesn't seem like a good idea. We should probably consider figuring out how to safely remove this:
https://github.com/RMI-PACTA/r2dii.plot/blob/main/R/utils-pipe.R

cjyetman:
- resolves #583


MonikaFu (approved):
LGTM. I run all the examples locally and everything is working as expected.

PR #610, opened 2025-03-25
Use "PACTA" instead of "2DII" in documentation

### LINKED ISSUE #604: Package website still mentions 2DII
First sentence when opening https://rmi-pacta.github.io/r2dii.plot/

> The goal of r2dii.plot is to help you plot 2DII data in an informative, beautiful, and easy way.

Change 2DII to `r2dii.analysis` output data or PACTA data? What would be more suitable, @jacobvjk @jdhoffa @cjyetman ?

MonikaFu:
Closes #604


R/theme_2dii.R
```diff
@@ -1,6 +1,6 @@
 #' Complete theme
 #'
-#' A ggplot theme which can be applied to all graphs to appear according to 2DII
+#' A ggplot theme which can be applied to all graphs to appear according to the PACTA
```
cjyetman, line 3: Nit: I'm glad to see this being updated, but wondering whether this misaligns the documentation/descriptions from the code, for instance here referring to "PACTA plotting aesthetics" for a function named `theme_2dii()`? Not a problem technically, but maybe causes some unnecessary confusion.

MonikaFu: I understand your concern. However, I think that since all functions in the `r2dii` suite use the `2dii` suffix it should not cause too much confusion to use it in the styling functions. At the same time, using the 2DII name as a name of the organisation should be terminated in my opinion. So I am willing to make this change even if it might cause some confusion.

MonikaFu:
@jdhoffa @cjyetman It's been a while since I've updated a package website but if I remember correctly, it will be handled automatically on GH once this PR is merged in?

cjyetman:
> @jdhoffa @cjyetman It's been a while since I've updated a package website but if I remember correctly, it will be handled automatically on GH once this PR is merged in?

yes, that's correct... though be aware that because the version number is a dev version number (0.5.1.9005), it will update the dev version of the site (https://rmi-pacta.github.io/r2dii.plot/dev) not the release version of the site (https://rmi-pacta.github.io/r2dii.plot/)

cjyetman (approved):

MonikaFu:
> > @jdhoffa @cjyetman It's been a while since I've updated a package website but if I remember correctly, it will be handled automatically on GH once this PR is merged in?
> 
> yes, that's correct... though be aware that because the version number is a dev version number (0.5.1.9005), it will update the dev version of the site (https://rmi-pacta.github.io/r2dii.plot/dev) not the release version of the site (https://rmi-pacta.github.io/r2dii.plot/)

Thanks! I remembered it was something like that!

